Issue #207, opened 2014-11-29
Where to put frappuccino?


mbrookes:
https://github.com/steveklabnik/frappuccino

Maybe misc? (Or may not at all...)

PR #289, opened 2015-01-17
Add Deterministic

Erol:


markets:
Hi @Erol, I think is a little bit esoteric and non-common used by the community. Not sure if we should add that kind of links to this project.

Erol:
Hmmm you have a point. This gem implements monads, which is probably more familiar to a Haskell developer. I'll close this for now.

markets:
Hi @Erol, about monads, I really like https://github.com/txus/kleisli by @txus.

We decide something similar (too obscure, esoteric; not really common used) there :

* https://github.com/markets/awesome-ruby/issues/207
* https://github.com/markets/awesome-ruby/pull/208

Thanks!

PR #291, opened 2015-01-17
Add ferry

amerov:
A ruby gem for easy data transfer for Rails app.


markets:
Nice one! Thanks :rocket:

PR #292, opened 2015-01-17
Add Trailblazer

amerov:


markets:
:+1: :100: I like a lot the idea behind Trailblazer!

PR #297, opened 2015-01-23
Add reference to daru

v0dro:


markets:
Hi @v0dro, thanks for your suggestion but I'm not sure if we should add this resource, I suspect isn't a very well known and used by the community. The idea is to keep in this list only best/well-known/MVP Ruby projects, otherwise can lose value and the list would be endless.

Sometimes is not easy to draw the line... in those cases we leave the PR open (if you think it's necessary) for some time and wait for people feedback.

This "Scientific" section was extracted in a row from https://gist.github.com/agarie/272ef4967479fedac856, the idea was discussed here https://github.com/markets/awesome-ruby/pull/165

Thanks.

v0dro:
I am currently developing this gem myself to address the very real growing need of a data analysis solution in ruby.

Its under active development as you can see from the commits, and is getting more downloads every day. See it here: https://rubygems.org/gems/daru

markets:
Yes, it seems under heavy development, so probably it's a bit fresh to add it right now (not stable yet, as mentioned in contribution guidelines https://github.com/markets/awesome-ruby/blob/master/CONTRIBUTING.md#quality-standard).

So, IMO, we should leave the PR opened in order to track the status and wait some time to see if the project becomes an essential tool and really useful to the Ruby ecosystem. I review issues/PRs, at least, once per week. Ping me for relevant updates.

PD. Daru looks promising, good job :+1: Hopefully the project will become an awesome tool. Keep in touch!

markets:
Hi @v0dro, merged in https://github.com/markets/awesome-ruby/commit/7404c5262ab5c024395190e82dcd57f46cec52ed. Thanks! Keep working on this!

PR #298, opened 2015-01-23
add interpolation

v0dro:


markets:
Similar to https://github.com/markets/awesome-ruby/pull/297

PR #316, opened 2015-02-05
Add Pragmatic Segmenter and ProZ

diasks2:


markets:
Hi @diasks2, 

First of all, thanks for your suggestions. I think Pragmatic Segmenter could be added to this collection, but ProZ isn't really a well-known/used library...

In order to avoid these kind of situations, we should modify contribution guides adding something like: "1 resource per Pull-Request". Will be easy to merge/discard/stand-by suggestions.

So for now, IMO we should not merge this PR.

diasks2:
Ok, thanks. I understand. I'll close this and send a new pull request with just Pragmatic Segmenter.

PR #319, opened 2015-02-07
Adding NetSuite gems

iloveitaly:


markets:
Hi @iloveitaly,

Thanks for sharing! 

I'm not sure if it's a really well-known/used/battle-tested library by the Ruby community. In these situations, we keep the PRs open for some time waiting for community feedback (+1 or -1 would be enough).

Also I think would be a good idea to submit just 1 project per PR => easy to merge/discard new suggestions.

iloveitaly:
Awesome, thanks for the feedback! I'll keep that in mind for the future.

markets:
I think I should modify contribution guidelines with that :sweat_smile:

Thanks! Sorry, but at that point, we already have a big list, IMO it's better to take care of what projects  we add to it, otherwise it may lose value. 

Let's build an awesome/useful collection for Ruby :rocket:

PR #329, opened 2015-02-19
Added backburner in queue section

diegobernardes:


markets:
Hi @diegobernardes, could you please add that resource in alphabetical order (as per contrib. guides)? Thanks! :+1: 

I think it's supported by ActiveJob, isn't?

diegobernardes:
Sorry about the order of items :P
And yes, ActiveJob supports backburner.

PR #331, opened 2015-03-04
add lina

huobazi:


markets:
Hi @huobazi, 

Thanks for sharing!

A couple of comments:

* Seems a bit fresh for now.
* Docs only in english?
* Links should be sorted alphabetically (as per contrib guides).

markets:

PR #334, opened 2015-03-10
Elastics gem

printercu:


markets:
Why this is better, or what are the advantages, versus the official Ruby SDK?

printercu:
It's much simpler, [more 'ruby/rails-way'](https://github.com/elastic/elasticsearch-rails/tree/master/elasticsearch-model#the-__elasticsearch__-proxy) and there is no strange code like [this](https://github.com/elastic/elasticsearch-rails/blob/master/elasticsearch-model/lib/elasticsearch/model.rb#L91-L103) (legacy?). I suppose it should be faster, but haven't benchmarked yet. There are also some features which are not available in official gem like migrations.

markets:
Interesting... Could you please rebase current master? I'd be happy to merge this!

printercu:
Rebased. Thanks!

markets:
Thanks :rocket:

PR #336, opened 2015-03-14
Add github_api

fpgentil:
A Ruby client for the official GitHub API
https://github.com/peter-murach/github
https://rubygems.org/gems/github_api


markets:
Hi @fpgentil,

We already have Octokit in that section (some lines below), the official Ruby client to interact with the Github API v3. 

Not sure if we should to add this gem, just used Octokit, so don't know if it's better in some points (code, api, tests, performance, maintenance, docs ...) or perhaps it introduces new/different features/approaches.

Let's keep this open and wait for feedback.

fpgentil:
Hello @markets 
I'm pretty sure I had previously searched for a gem to use with Github API, thanks for pointing that out! :)

I'll try to dig into both codes and see if there's anything different in terms of features/approaches that might be worth to have them both in this project, otherwise I agree with you.

markets:

PR #341, opened 2015-03-23
Add Opal

Erol:


markets:
Hi @Erol,

I merged this, but not sure about category. I'm thinking that would be nice to add a new category with the different Ruby "implementations": MRI, JRuby, Rubinius, Opal...

Thoughts? Probably top-level category (awesome-ruby, services and apps, resources)?

PD. I really like Opal, sometimes I think that I should start use Opal instead of CoffeeScript :sweat_smile:

Erol:
How about **Ruby Implementations/Compilers**? (Since Opal is a Ruby to JS compiler) :)

markets:
`Ruby Implementations/Compilers` :+1:

PR #344, opened 2015-03-24
Move Opal under Implementations/Compilers

Erol:
In line with #341, I'd like to introduce another category called **Implementations/Compilers** where compilers like Opal and implementations like Rubinius & JRuby can be placed under.

What do you think about this @markets?

Thanks!


markets:
I'm totally :+1: on this!

PR #378, opened 2015-04-08
Add reference to Surrounded

saturnflyer:
I actively use and maintain Surrounded to handle separating business logic from the objects that need it. It is well documented in the readme explanation and examples of use, description of implementation, and instruction on handling alternative behavior https://github.com/saturnflyer/surrounded/blob/master/README.md


markets:
Nice one! Thanks :rocket:

PR #380, opened 2015-04-09
Add Turbolinks

theareba:


Ch4s3:
Turbolinks ships with Rails, so I think it's pretty visible already.

markets:
@Ch4s3 agreed, also it's mostly a coffeescript project (Ruby code is just to load into Rails)

PR #389, opened 2015-04-10
Add a reference to validate_overlap

supertinou:
## Add a reference to validate_overlap

__When this gem should be helpful for you?__

If you are developing Rails 3 app, let say some meeting planner and you can't save records which have time overlap.


markets:

Issue #304, opened 2015-01-28
Small gems


obduk:
Hi, first of can I say this is a very useful page.

One problem I have found with it is a lot of the gems on the list have very few download numbers, some even below a hundred. Although I don't want to discourage new gems, I often find gems with so few download numbers are either not properly maintained, or are not greatly peer reviewed.

Would it be a good idea to have a rough download count next to each one, or a tag for new gems?

mbrookes:
I agree with the concern, but adding counts to each listing and keeping it
updated would be very time consuming as the list is manually curated.

I tend to add useful looking, well structured gems but which are new or
with low star count as issues rather than PRs.

Some guidance in the contributing read me on minim age and star count might
be a good idea, but then you can have old abandoned gems that were popular
in their time still qualify, so it wouldn't fool-proof.

Tangentially, I also think there's a case to be made for splitting out
Rails-only and generic Ruby gems into separate lists.

>

markets:
Hi @obduk,

I also agree with the concern, but as @mbrookes said, isn't so maintainable. After all, this is just a markdown file (=> parsed via GitHub API to build a site) + issues. 

We started discussing this idea in: https://github.com/markets/awesome-ruby/issues/60.

I think `Awesome Ruby` should be a place to find most awesome Ruby related projects, bests tools to build/test/debug/analyze/deploy/package... Ruby apps nowadays and really really useful in general. That context also includes new gems (if they fit contribution guides). 

The idea is to collect MVP resources per category, but once here (a category scope), the idea is to explore/review those top filtered tools by oneself. Numbers are usually good indicators, but sometimes awesomeness isn't just numbers...

> Some guidance in the contributing read me on minim age and star count might
be a good idea, but then you can have old abandoned gems that were popular
in their time still qualify, so it wouldn't fool-proof.

Probably would be a good start. For example: stars + watchers > 100, active last year. Of course, not all projects are hosted on GitHub, so for those cases and for other exceptions we can use GH issues to discuss about it. 

Sometimes I think in moving the project to static site backed with YAML, would be nice to have quantitative measures, metadata, categories, tagging... and have an script to recompute data via APIs, just an idea right now... ++maintainability

Another idea: write a script (aka `awesome-ruby-doctor`), run it locally to detect low level resources, in terms of stars, downloads, activity... would be really helpul to maintain that list in an easy way.

markets:
Ah @obduk, what gems are you talking about? 

[...]

Issue #60, opened 2014-07-12
Voting system


markets:
In order to ensure higher quality of resources, could be a good idea to introduce some kind of simple voting system.
- 5 postive votes (+1) => add it to the list (or merge PR if any)
- 5 negative votes (-1) => remove it from list

What do you think? We can add a VOTING.md file explaining the rules or just add a section into the CONTRIBUTING.md document.

markets:
Let's use GitHub reactions and comments to let the community to express their thoughts. Added to contribution guides here https://github.com/markets/awesome-ruby/commit/56ecd65d0d7c221a0103344b2dfc0d2e2eb65787

bernsie:
would it be feasible to implement github's [`.github` folder templates](https://github.com/blog/2111-issue-and-pull-request-templates) to enforce these guidelines?

markets:
hi @bernsie, do you mean to copy the content of CONTRIBUTING.md file to a PULL_REQUEST_TEMPLATE.md? Would be nice if GH allow symlinks for that... 

PS. this works right now https://github.com/blog/1184-contributing-guidelines

bernsie:
something along those lines yes.

markets:
@jbernsie pushed [here](https://github.com/markets/awesome-ruby/blob/5a43340d7e2ccacddcb03938fea1a33b5523efba/.github/PULL_REQUEST_TEMPLATE.md), what do you think about content (or formatting)?

jbernsie:
@markets Looks good to me man 👍 no complaints haha

PR #392, opened 2015-04-10
Add proxy_pac_rb to testing

maxmeyer:


markets:
Thanks for sharing @maxmeyer. This gem looks nice, but I think we can't consider it as a well-known resource. We should add only well-known and popular used tools to keep some value on this list.

maxmeyer:
> This gem looks nice, 

Thanks a lot. :smile:

> well-known resource. We should add only well-known and popular used tools to keep some value on this list.

Mmh... I thought the list is kind of a source for finding "diamonds" - I avoid the word gem here to prevent misunderstandings - something like I have problem x and want to find a gem which solves my problem.

No offense, but what's the worth of having five different gems for commandline parsing?

Maybe it does make sense to add this `We should add only well-known and popular used tools to keep some value on this list.` to the CONTRIBUTIONS.md then?

maxmeyer:
Or maybe it does make sense to add something like... you need a least "20" stars to be interesting for the list.

markets:
Hi @maxmeyer, first of all thanks for your feedback, much appreciated :+1: 

We also discussed this topic here: https://github.com/markets/awesome-ruby/issues/304#issuecomment-72120807.

It seems would be a good idea to add this quantitative measures to the contrib guides, ie: stars+contributors > 50, active last year and >10k downloads.

Sometimes it's not easy to draw the line... but if we add all resources nor will help :shipit:

A kind of voting system (via GH issues https://github.com/markets/awesome-ruby/issues/60) probably could help to maintain the collection.

markets:

PR #396, opened 2015-04-10
Add proxy test endpoint

maxmeyer:
This is a rails application which can be used as end point for your test suite. It is designed to cover the most problematic areas when you run proxies in a large corporate environment.


markets:
Similar to https://github.com/markets/awesome-ruby/pull/392

PR #397, opened 2015-04-10
Add filegen to library

maxmeyer:
You can use filegen when you need to generate files based on environment variables. I use this application when I build rpm packages via an internal [open build service](http://openbuildservice.org/).


markets:
I think this isn't generally useful to the community. It covers a very specific use-case.

maxmeyer:
Mmh... What a bummer...

PR #399, opened 2015-04-11
Added sup

maxmeyer:
this is ruby bases mailclient with a nice and active community.

@gauteh: Just for you to know.


gauteh:
:beers:

PR #403, opened 2015-04-16
Add WialonApi

thorn:


markets:
Hi @thorn, thanks for sharing!

But we can't consider it a well-known resource (1 star, 800 downloads) in the Ruby ecosystem.

In these situations we keep the issue open for some time in order to track if the resource gains popularity. You can see similar cases in the issues/PRs lists. Thanks!

PR #405, opened 2015-04-20
Add active_link_to

GBH:
https://github.com/comfy/active_link_to


derek-watson:
:+1: Great gem, I use it in every project

PR #407, opened 2015-04-28
Add `acts_as_api` to API Builders

esbanarango:


markets:
Hi @esbanarango, thanks for sharing!

It doesn't seems very active, last commit on `10 Mar 2014` (more than 1 year). Do you think is actually a good way to go for APIs?

esbanarango:
Hello @markets.

We use this on production in multiple projects. It's not actually abandoned but done/ready. It has a really useful feature which is "Supports multiple api rendering templates per model. This is especially useful for API versioning or for example for private vs. public access points to a user’s profile". That for example is missing on _ActiveModel::Serializers_.

I would definitely recommended.

markets:
Hi @esbanarango,

I just did a quick review and yes, it has different ways/approaches to build APIs.

But I think it doesn't support for example, Rails 4 (version 5 is on the horizon) and Ruby 2 (https://github.com/fabrik42/acts_as_api/blob/master/.travis.yml). Not sure if it's appropiate for new projects.

Let's keep the issue opened and wait for community feedback.

Thanks!

markets:

Issue #196, opened 2014-11-22
active_type gem


woto:
Here is another implementation of Form/Service Objects (not sure in category)

[active_type] (https://github.com/makandra/active_type) - Make any Ruby object quack like ActiveRecord

markets:
Closing for now, please re-open in case you want to resume that discussion. Thanks!

PR #415, opened 2015-05-07
add ActiveType

stamm:


markets:
hi @stamm, a while ago we discussed a bit about it... https://github.com/markets/awesome-ruby/issues/196

stamm:
@markets but it present not in list at now. What I need to do: move to that category?

markets:
I'm not sure if we should add this resource, in this case `ActiveAttr` seems, by far, a more popular (stars, downloads, contributors...) and widely used to achieve the same thing.

In addition to inheritance vs mixin, are they significantly different in other ways?

stamm:
It seems, you're right, I'm not go deep to compare this gems

markets:

PR #416, opened 2015-05-07
Add Derailed Benchmark

fpgentil:
Derailed benchmark is some sort of collection of other gems to perform a benchmark. I wasn't sure if it should fit here as it's focused on Rails, but you can also use with Rack so I gave it a shot :smile: 

```
url: https://github.com/schneems/derailed_benchmarks
description: Go faster, off the Rails - Benchmarks for your whole Rails app
```


markets:
So, what do you think if we modify a bit the description to be clear about Rack compatibility?

fpgentil:
It makes sense, I'll update the PR..

fpgentil:
What do you think, @markets? :smile:

markets:
Thanks! Now it's more precise :+1:

PR #418, opened 2015-05-11
Add Trollop with link to updated github page.

guihaojin:
I have updated the link to @kbrock's Trollop page. Please check if it's valid, thanks!


kbrock:
@guihaojin Thanks so much

markets:
Thank you a lot @guihaojin @kbrock 

This https://github.com/wmorgan/trollop/pull/2 would be helpful for future visitors :+1: 

Thanks again @kbrock for that and for retrieve Trollop :sunrise: (would be also useful to annouce that in other active Ruby communities, for example Reddit [r/ruby](http://www.reddit.com/r/ruby/)).

PR #421, opened 2015-05-18
Add chamber to the Configuration list

jfelchner:


markets:
Thanks @jfelchner! I recently tried chamber and I found it awesome, then I remembered that there was an open PR, so time to 🚀

jfelchner:
@markets better late than never 😂  Glad you loved it!

PR #426, opened 2015-05-27
Honeybadger is an error handler at heart.

bragboy:
Honeybadger is basically an error reporting mechanism similar to Airbrake. I have used both of these applications and both these applications' main objective is to capture and aggregate run time errors. In that aspect it is right to list Honeybadger under `Error handling` instead of the miscellaneous `Services and Apps` section.

Please check the website for proof: https://www.honeybadger.io/

image


markets:
Hi @bragboy, thanks for commenting. If you review the links under `Error handling`, you'll find only libraries to deal with errors (gems) and ruby-clients (gems) to other services (Airbrake, Raven, ...). However, in `Services and Apps` we actually list services (SaaS). So I think this is the proper place for a resource pointing to https://www.honeybadger.io.

It probably would be helpful to split a little the `Services and Apps` section adding new categories: Monitoring, CI, Docs, Management, ...

jfelchner:
@markets just my opinion but I agree with @bragboy.  Honeybadger is definitely in line with Airbrake and whichever section they're in, they should be in the same one.

* Raygun
* Bugsnag
* Exceptional
* Honeybadger
* Airbrake
* Errbit

They should all be in the same place.

PR #427, opened 2015-05-28
Add Monit

hbin:
Monit is a small Open Source utility for managing and monitoring Unix systems. And it's premium version M/Monit expand on Monit's capabilities and provides monitoring and management of all your Monit enabled hosts via a modern, clean and well designed user interface which also works on mobile devices.


markets:
Hi @hbin, thanks for sharing. Monit is awesome, but is written in C (I think). So it isn't related specifically with the Ruby ecosystem.

hbin:
Oops, I didn't realize that. I got it from [Railscast#375](http://railscasts.com/episodes/375-monit?view=asciicast).

I use Monit a lot. It's realy awesome. And I'm also a rubyist!!

I think, we shouldn't ignore it since it's not written by Ruby. [Passenger](https://github.com/phusion/passenger) is almost written by C++, C.

markets:
Yep, but Passenger is kind of specific app/web-server for Ruby apps (now also runs Python and Node), it's very popular in Ruby community and originally (`mod_rails` and `mod_rack`) designed for web applications built on Rails. Monit is a monitoring tool for any kind of process.

hbin:
Fine, thanks for your feedback.

PR #431, opened 2015-06-17
Add Hobbit Microframework

tak1n:


markets:
:+1: I found an interesting benchmark of Ruby web-microframeworks running **Hello World** like apps: https://github.com/luislavena/bench-micro by @luislavena. Thanks!!

It's curious to see how the `Hobbit` is almost 3 times faster (req/sec) and does a third of memory allocations compared to `Sinatra`.

tak1n:
I found `Hobbit` in this benchmark the first time too :+1:

PR #435, opened 2015-07-01
Update README.md

Gargol:
Removed `youtube_it` gem from the list as it only supports youtube api v2 which is deprecated and has been turned off on April 4th 2015. A better option is using 'yt' gem which supports the newest v3 api.


markets:
Thanks @Gargol :+1: :scissors:  :rocket:

PR #440, opened 2015-07-17
Add Rgl

Ch4s3:
rgl is a cool gem that gives you a lot of graph data structures and can produce pngs.


Ch4s3:
Thoughts?

markets:
Thanks @Ch4s3 !! :rocket:

PR #455, opened 2015-08-21
Adds validates_zipcode

dgilperez:
Hey!

What do you think of adding this little validation gem I put together? It's fresh, but it's already support for mostly every country format available, so it's stable.

Great work here btw!


markets:
Hey @dgilperez, thanks!!

As you said, it's currently a bit fresh. Anyway, it seems well designed, tested and documented, so it can has a place here :+1:

markets:
Closing for now, please re-open in case you want to resume that discussion (or the project reaches higher popularity). Thanks!

PR #457, opened 2015-08-24
Adding Vexor CI

makrusak:
Adding Vexor CI.
https://en.wikipedia.org/wiki/Vexor
https://vexor.io/

### DIFF
--- a/README.md
+++ b/README.md
@@ -1057,6 +1057,7 @@ Best suited for map-reduce or e.g. parallel downloads/uploads.
 * [Traveling Ruby](http://phusion.github.io/traveling-ruby/) - Traveling Ruby lets you create self-contained Ruby app packages for Linux and OS X.
 * [Travis CI.com](https://travis-ci.com) - Take care of running your tests and deploying your private apps.
 * [Travis CI.org](https://travis-ci.org) - A distributed build system for the open source community.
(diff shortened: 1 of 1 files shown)


makrusak:
Hello, what does "add link" label mean?

markets:
Hi @makrusak,

It's just means this PR contains a suggestion to add a link to the collection.  

I'm not sure if we should merge this (in reality, we should re-think this "Services" section)... so in these situations I add this label and wait some time comments (or +1, or -1) from community.

makrusak:
@markets.
1. Vexor can run unlimited number of simultaneous builds. When a lot of commits are submitted in a short time developers won’t have to wait in the queue. Also tests in each build are executed simultaneously.
2. You pay for the minutes of testing.
So developers aren’t afraid to make tests parallelized with Vexor because they shouldn't pay for idle servers.
As the result: faster testing, lower price.

makrusak:
So, @markets , what do you think about this pull request?

markets:
Didn't tried yet, but unlimited paralel builds + pay per minutes are really good features! :+1: 

Thanks @makrusak

makrusak:
@markets Maybe do you know any other places to tell about this to community?

PR #469, opened 2015-09-24
Add Camaleon CMS

Tvinky:
Camaleon CMS is an alternative to [Wordpress](https://wordpress.org/) for Ruby on Rails. Thought it would be good to have it listed under `CMS` category.


markets:
Thanks @Tvinky :+1: :rocket:

PR #470, opened 2015-09-27
Added any_login gem

igorkasyanchuk:
Added gem - https://github.com/igorkasyanchuk/any_login


markets:
What about https://github.com/flyerhzm/switch_user? Main dis/advantages?

igorkasyanchuk:
@markets when I was doing my gem I didn't know about switch_user, but even if I know about it I would develop my gem because it's more flexible and has more configuration options and I think easier to use. But this is something what users will decide :)

markets:
Yes, let's leave users to decide if this deserves a place in this list 👍 

Just some (interesting?) numbers:

- any_login: 216 stars, 7 contributors, 18k downloads
- switch_user: 637 stars, 30 contributors, 416k downloads

PS The `Misc` section was removed (https://github.com/markets/awesome-ruby/commit/7283550970942808f0c5d40cfbff99fa048b8248), in case we finally merge this, we should move it to a more specific category.

Issue #472, opened 2015-10-08
RailsBricks is over? 


markets:
* https://github.com/nicoschuele/railsbricks => 404
* http://www.railsbricks.net/

PR #473, opened 2015-10-09
Remove RailsBricks

fpgentil:
As mentioned here #472 

* http://www.railsbricks.net
```
I have left the Internet.
So Long, and Thanks for All the Fish
```

* https://rubygems.org/search?utf8=✓&query=railsbricks
```
NO GEMS FOUND
```


markets:
Thanks @fpgentil :+1:

PR #474, opened 2015-10-12
Added google-api-ads-ruby gem to Third-party APIs

kowal:
See https://github.com/googleads/google-api-ads-ruby.


kowal:
@markets what do you think about this?

markets:
Yep, thanks @kowal 🚀

PR #481, opened 2015-10-22
add ruby-advisory-db gem

mabras:


markets:
Hi @mabras, I think we don't need to merge this one, since we already have listed `bundler-audit` under https://github.com/markets/awesome-ruby#security. If I'm not wrong, `ruby-advisory-db` is the "db" consumed by the `bundler-audit` tool.

Thanks.

PR #483, opened 2015-10-23
Add a few gems I've found very useful.

musl:


markets:
Hi @musl, thanks for your suggestions! 

As per contribution guides, you should submit 1 link per Pull-Request, in order to facilitate discussion.

Anyway, those gems seems less popular, used and maintained compared to the other ones under same categories.

So, If you think they are really awesome (or better than alternatives already listed), we can review it deeply one-by-one, just take in account this comments... Otherwise this collection will lose value, the key (IMHO) is to only mantain most MVP Ruby (well-known, popular, used, generally useful, active, tested). Sorry sometimes it is not easy to draw the line...

Thanks.

musl:
No worries! Thank you for the consideration.

PR #486, opened 2015-10-27
add the `combine_pdf` and `plezi` gems

boazsegev:
CombinePDF had been gaining traction and allows PDF manipulation without the need to install external libraries or helpers such as `iText` or `pdftk` - this allows CombinePDF also to work entirely in memory, parsing and outputting PDF data without using system calls or requiring temporary files.

Plezi is not as well known and is quite new to the playground, but it has an edge that makes it very powerful - it scales easily using Redis, allowing broadcasting and unicasting between Websockets to work seamlessly across different processes with no shared memory (i.e. having many Dynos on Heroku).

Plezi's Object Oriented model and Websocket API makes it reminiscent of the early days when writing for Rails was super simple. It makes writing Websocket applications that easy.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1024,7 +1024,6 @@ Best suited for map-reduce or e.g. parallel downloads/uploads.
 * [Lotus](http://lotusrb.org) - It aims to bring back Object Oriented Programming to web development, leveraging on a stable API, a minimal DSL, and plain objects.
 * [Padrino](http://www.padrinorb.com) - A full-stack ruby framework built upon Sinatra.
 * [Pakyow](http://pakyow.com/) - A framework for building modern web-apps in Ruby. It helps you build working software faster with a development process that remains friendly to both designers and developers.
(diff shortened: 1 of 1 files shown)


markets:
Hi @boazsegev, thanks for sharing these nice projects!

`CombinePDF` :+1: 
`Plezi` looks promising, but a bit fresh for this collection at this moment... someone using it in production?

Please, don't forget to review the contribution guides :book: => one suggestion per PR. 

So in order to move this forward, I think you could update this PR adding `CombinePDF` only.

boazsegev:
Hi, Thanks so much, both for the feedback and the amazing compilation 👍🏻

As for the contribution guide... yeah, my bad, sorry. I somehow missed it, getting lost in the list of amazing gems.

I removed the references to the Plezi framework for now, because I have no idea if anyone is using it in production. I do know I got some limited feedback, but it could have been people testing it out... time will tell and when the time comes I might ask you about it.

PR #487, opened 2015-10-27
add NoBrainer orm

Ch4s3:


markets:
Hi @Ch4s3, :+1: for this, but If I'm not wrong, this isn't only Rails. Should be something like:

    NoBrainer - A RethinkDB ORM for Ruby

Also, I don't see the need to add a link to RethinkDB.

Ch4s3:
@markets noted, and changed.

markets:
Thanks :shipit:

PR #492, opened 2015-10-30
Adding IFTTT/Kashmir

nettofarah:
Adding it in two categories because it can be used for both caching and object serialization (for APIs).


markets:
Looks nice IMO, but we should decide between serializers or caching ... add a resource to a different sections won't help to mantain this list, this situation could happen more times, and start duplicating resources it's not a good idea. 

PS. Please, don't forget the alphabetical order.

nettofarah:
Alright. Thank you, @markets.
It should be good now.

markets:
Hi @nettofarah, 

I think there is something wrong yet (`snap.` on next line), could you please review it? 

PS. If you can squash to a single commit would be cleaner...

markets:
ping @nettofarah

nettofarah:
Sure, @markets. I'll have access to a computer tomorrow morning.
I'll get this squashed and pushed!
Thank you!

nettofarah:
We should be good now, @markets. Sorry for the delay.

markets:
Thanks @nettofarah :rocket:

nettofarah:
yay!

PR #498, opened 2015-11-08
Remove Campo

jjohnson:
Campo readme states project is inactive.


markets:
:+1: thanks! :rocket:

PR #502, opened 2015-11-11
Add Pry Byebug

jjohnson:


markets:
Hi @jjohnson, thanks! I think we can remove `Pry Debugger`... it targets only Ruby < 2. Also, we can probably move `Pry Byebug` under https://github.com/markets/awesome-ruby#debugging-tools. What do you think?

jjohnson:
I agree. Moved under debugging-tools. Created #503 to remove debugger.

markets:
Thanks a lot @jjohnson :rocket: :shipit:

PR #505, opened 2015-11-13
Add Naught to the Core Extensions list.

nilcolor:


markets:
Hi @nilcolor, I really like most of @avdi stuff (Ruby Tapas is👌), but I think we can't consider this gem as an essential tool to build Ruby apps. I like the pattern, but I prefer to build this kind of things with plain Ruby (no deps).

PR #512, opened 2015-11-28
tiny spelling fix; rouges to rogues

CameronCarroll:
Almost didn't even want to make a request for something so small, but it jumped out at me.


markets:
Thanks @CameronCarroll :+1:

PR #515, opened 2015-12-04
Add Lotus::Utils to Core Extensions

joneslee85:


markets:
Hi @joneslee85, sorry for the delay...

A couple of comments:

* Did you know if someone is using it outside Lotus? 
* I can't find specific documentation about Lotus::Utils.

I really like Lotus and the architecture/modularization behind, but I'm not sure if its components are used in no-Lotus apps nowadays.

Anyway, Lotus::Utils seems the most downloaded component: https://rubygems.org/search?query=lotus-

Thanks! Keep in touch.

joneslee85:
@markets thanks for the reply. I myself use this in a sinatra app and I know few other projects from friends who also replace ActiveSupport with it.

I acknowledge your concern about Lotus gems, please let me assure you that every components of Lotus were designed with simplicity and independence in mind. That means it is out of the box recommended to be mixed into other app / frameworks. Lotus Framework itself, alone is just a glue-component that bind all components together, very loose in fact. This whole concept that makes Lotus lightweight and customisable for different usecases

joneslee85:
@markets regarding API, please look at http://rdoc.info/gems/lotus-utils

markets:
Ok @joneslee85, thanks for the explanations! 

In that case, it make sense to add some of them. I'd say, the most used outside Lotus or, at least, with more potential to replace parts in other Ruby apps:

* `Lotus::Utils` ~> `Active Support`
* `Lotus::Model` ~> `Active Record`

PR #519, opened 2015-12-05
Semantic Logger and Symmetric Encryption have moved

reidmorrison:
Repos moved to a Github organization to facilitate shared ownership and maintenance


markets:
Thanks @reidmorrison :rocket:

PR #520, opened 2015-12-07
Add Runfile to CLI Builder

DannyBen:
https://github.com/DannyBen/runfile


markets:
Thanks for sharing it @DannyBen 

But we can't consider it a widely-used/top/mature resource in the Ruby community. Seems a bit fresh for now.

PR #521, opened 2015-12-07
Add docopt to CLI Builder

DannyBen:


markets:
Looks a bit abandoned no? Last commit was on `1 Sep 2012`, no answers in issues ...

DannyBen:
docopt as a whole, seems [very much alive](https://github.com/docopt/). I agree the Ruby port seems to be unmaintained at the moment, but still, this is by far the easiest way to create Ruby CLI in my opinion.

markets:
Yes @DannyBen, as a whole project, seems totally alive, but specifically, the Ruby version seems not well maintained. I think we have pretty good solutions under `CLI Builder` and better supported by the teams behind. So, to start a new project, I would not recommend to use an "unmaintained" dependency.

Let's wait for other Rubysts feedback, but at first, I'm :-1: on this.

markets:
Sorry @DannyBen, from my point of view, this repo seems actually unmaintained.

PR #524, opened 2015-12-15
Added a couple more "feature flippers"

joelvh:


markets:
Hi @joelvh, thanks a lot for sharing! But, as per contribution guides: 

> Add one link per pull-request.

This way it's easy to add/reject projects individually :+1:

raysegantii:
:+1: How about https://github.com/jnunemaker/flipper as well?

markets:
hi @raysegantii, agreed :+1: I used Flipper in the past and it works quite well. 

So, in summary, we have 4 proposals for flippers, we should pick just the best options nowadays in terms of: functionality, maintenance, popularity, community...

IMHO: `amatsuda/motorhead` and `jnunemaker/flipper`. What do you think guys? What are the "pros" of `balvig/chili` and `pda/flip` @joelvh ?

PR #525, opened 2015-12-17
add rails-i18n gem

chungphuoc:
https://github.com/svenfuchs/rails-i18n


markets:
Hi @chungphuoc, thanks! Btw, I see a couple of things:

* Bad format
* Not alphabetically sorted (as per contribution guides)

Could you please address those things?

jdickey:
Also, the URL in the repo's description isn't available (http://rails-i18n.org/ displays a parked-domain page for "Strato"). You might want to either get the site up *or* remove the link from the description until you do have it up.

chungphuoc:
The domain maybe expired but the gem https://github.com/svenfuchs/rails-i18n is helpful for multiple languages in active records errors. You should take a look

markets:
Thanks for that point @jdickey! I see a related issue in the repo https://github.com/svenfuchs/rails-i18n/issues/603

n-rodriguez:
Hi there! Any news? 9 028 115 downloads... I do think it's used ^^

markets:
you're right @n-rodriguez! still needs some work to be merged this https://github.com/markets/awesome-ruby/pull/525#issuecomment-165787591

n-rodriguez:
> still needs some work to be merged, see this comment

Yeah... but I don't think this will happen... the issue is almost 3 years old... I can provide a "Fix typo" commit if you like :)

markets:
That would be amazing @n-rodriguez

PR #535, opened 2016-01-11
Add SecondBase Gem

metaskills:
Seamless second database integration for Rails. SecondBase provides support for Rails to manage dual databases by extending ActiveRecord tasks that create, migrate, and test your application.

https://github.com/customink/secondbase


metaskills:
cc @hmadison

markets:
Nice one @metaskills! :rocket: :+1:

PR #544, opened 2016-02-03
Add scenic gem

teerawat1992:
Add scenic https://github.com/thoughtbot/scenic


markets:
thanks @teerawat1992 :rocket:

PR #546, opened 2016-02-04
Add scientist gem

teerawat1992:
Add scientist gem https://github.com/github/scientist


markets:
Nice one @teerawat1992! thanks :+1: :rocket:

PR #551, opened 2016-02-14
Add attr_extras

hothero:


markets:
Hi @hothero, thanks for sharing this resource.

Since this is just some boilerplate, I'm not sure if we can consider it essential Ruby. Other resources in that section (Virtus, ActiveAttr and FastAttributes) are doing quite more things.

PR #556, opened 2016-02-24
Add link to skinny_controllers

NullVoxPopuli:


markets:
Hi @NullVoxPopuli, thanks for share this resource!

Is this similar to https://github.com/josevalim/inherited_resources?

NullVoxPopuli:
Not entirely. I'd say it's a cross between that an trailblazer

markets:
I'm asking this, because Inherited Resources is kind of deprecated...  but yes, this project also offers different stuff (Operations). Actually it seems a bit fresh for this collection (just 5k downloads, 50 stars ...), but we can keep the issue open (as we usually do) and let the community comment.

NullVoxPopuli:
For what it's worth, based on what I saw in Inherited Resources, the goal seems to be that you just don't have to define any actions in your controller.

It would be really straight forward to add that to skinny_controllers.

```ruby
module SkinnyControllers::DefaultController
  include SkinnyControllers::Diet

  def index
    render json: model
  end

  def create
    render json: model
  end

  def update
    render json: model
  end

  # ... etc
end
```

and then in a controller, someone could juste `include SkinnyControllers::DefaultController` to get all their actions fully defined.

markets:
But they deprecate it for that reason: to have no actions defined in a controller it's usually too much magical and implicit, it's probably suitable for prototyping apps or for really simple CRUDs apps ... 

Your project is pretty different in that aspect, it's just straight forward to "simulate" this behaviour (as per your code snippet), but nothing enforces you to do it :+1:

PR #571, opened 2016-04-04
Blanket should be categorized in HTTP, Not API Builder

AknEp:
Blanket is not a "API Builder", but API Wrapper.
I think "API Builder" list should contains only API builder (≒ server).

### DIFF
--- a/README.md
+++ b/README.md
@@ -142,7 +142,6 @@ Thanks to all [contributors](https://github.com/markets/awesome-ruby/graphs/cont
 ## API Builder
 
 * [ActiveModel::Serializers](https://github.com/rails-api/active_model_serializers) - JSON serialization of objects.
-* [Blanket](https://github.com/inf0rmer/blanket) - A dead simple API wrapper.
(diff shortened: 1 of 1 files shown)


markets:
Thanks @AknEp, you're right! It's kind of "API explorer"... what do you think if we change the category name? `API Builder` => `API Builder/Explorer` or just `API` ... or something similar. I think Blanket fits better under a category related to APIs ... more than `HTTP`. Thoughts?

AknEp:
@markets oops, sorry for my slow reply!

I think `API Builder/Explorer` is better than `HTTP`, and separate `Builder` and `Explorer` is better.

example:

> ### API
> - Explorer
>   - Blanket
>   - RestClient
>   - ...
> - Builder
>   - Grape
>   - Rails::API
>   - ...

As you know, `API` is originally means `application interface`, but I think `API` is used to describe about HTTP(S) (Rest) API, specifically in ruby/rails community.
And some of api-wrappers can wraps not only http but also ftp or another protocol.

PR #580, opened 2016-04-10
Add Eldr to Web Frameworks

k2052:


markets:
Hi @k2052, thanks for sharing this resource!

Do you know if any team is using Eldr in production? According to RubyGems it has a few downloads (under 2k).

k2052:
@markets not that I know of. I have used it on a few production projects myself. That said, I definitely wouldn't consider it a production ready framework. It is definitely more of a learning resource and a toy project. Is _production ready_ one of the criteria you use when including a framework on the list?

markets:
Hi @k2052, not like a "unbreakeable" rule, but in general I'd say yes. As per [Quality standard](https://github.com/markets/awesome-ruby/blob/master/CONTRIBUTING.md#quality-standard):

> Generally used and useful to the community.

So, let's keep this issue opened for some time (as we usually do) and see how it evolves. Also, some community feedback, or personal experience, is really helpful in this situation.

PR #581, opened 2016-04-11
Add Passages Ruby on Rails Engine

yez:
`passages` is a Ruby on Rails Engine aiming to help alleviate some issues
  with cumbersome Ruby on Rails Routes ([More information here](http://jakeyesbeck.com/2016/01/03/passages/))


markets:
Hi @yez, this one seems a bit fresh for that list (~ 2k downloads). On the other hand, since it's an extension to an already existing feature in Rails, but improved... not sure if we should these kinds of gems... the list can grow a lot, don't you think?

yez:
Hi @markets, what you said makes sense. Maybe if it changes to be a more popular Engine I can resubmit it. Closing for now, thanks for the feedback!

PR #583, opened 2016-04-11
Add Compel gem

joaquimadraz:
[Compel](https://github.com/joaquimadraz/compel), coerce and validate ruby objects.


joaquimadraz:
Hey @markets, do I need to add more info?

markets:
Hi @joaquimadraz, thanks for sharing! It's a bit fresh for now, under 3.7k downloads... we can't consider it a widely used project by the Ruby community... 

That said, nice project 👍 , I have recently seen several libraries regarding attributes coercion. Did you see the work under http://dry-rb.org? 

If we finally merge this one, I think it fits better under `Core Extensions`, next to Virtus and other attributes/objects extensions, don't you think?

joaquimadraz:
Hey @markets!

I found out about dry-validations couple of hours after I published the first release of compel :)
I find that the dry-rb DSL's can become confuse.
I really like [joi](https://github.com/hapijs/joi) and [React.PropTypes](https://facebook.github.io/react/docs/reusable-components.html) way of creating simple/explicit coercion and validation.

Yup, it makes sense on `Core Extensions`, I will update later then.

Thanks for the feedback, really appreciated 👍

PR #588, opened 2016-04-14
Add AccessGranted

pokonski:
https://github.com/chaps-io/access-granted


markets:
👍  thanks @pokonski nice one 🚀

Issue #384, opened 2015-04-10
Adding github stars


bertomartin:
It's still hard to decide which gem to choose (from each category) by browsing the awesome ruby website. It would be nice to see the number of github stars next to the gem.

markets:
Hi @bertomartin,

We already start to discuss a similar topics/ideas here: https://github.com/markets/awesome-ruby/issues/304, https://github.com/markets/awesome-ruby/issues/189

Some quantitative data would be useful, but the main goal of this project isn't to collect most starred repos... After all, this is just a markdown file.

Also, projects under a category are, sometimes, complementaries and they provide different features/approaches.

Issue #189, opened 2014-11-13
Data YAML/JSON backed and site redesign


markets:
Continuation of some thoughts in https://github.com/markets/awesome-ruby/issues/60#issuecomment-61476505

**Summary**: Move resources (links) to YAML (instead of Readme) with more metadata (fetch via apis: stars, last commit, downloads ...) and build the site via scripts. Still easy to contribute, powerful static site => better access/filter info. Just an idea to think about, any kind of suggestions are really welcome. 

Another good example of JSON data backend + tagging: http://googlecloudplatform.github.io/

@AlexLarra started a simple script to convert the Readme to a YAML: https://gist.github.com/AlexLarra/0ac306bfae319a940909

maxmeyer:
@markets 
Maybe you want to look at https://github.com/middleman/middleman-directory there's a [data file](https://github.com/middleman/middleman-directory/tree/master/data) where information is held for `middleman-extensions`

@tdreyno

Would it be easy to fork and re-use the directory for another project? Any suggestions?

tdreyno:
@maxmeyer You're welcome to fork/copy/change whatever helps you out. We're probably going to rebuild the directory in the latest Ember release at some point in the future.

markets:
Thanks @maxmeyer @tdreyno !!

I recently found that project and this is exactly the idea I had in mind (if we finally take the decision to rewrite AwesomeRuby). I also like: http://googlecloudplatform.github.io/ and http://motion-toolbox.com/ (middleman project).

Would be nice to extract main funtionality of middleman-directory to a middleman extension (released as a gem). Lately, there is a lot of directory/collection (organized by tags) projects and I suspect most of them would use this extension :100:

maxmeyer:
@markets I'm not sure if this - the extension - makes sense. I think most logic is part of the ember.js-logic, correct @tdreyno? Maybe you can use `bower` to build a re-usable component. If you need help on the ruby-stuff I think I can help, but I'm definitively no JavaScript-expert. :smile: 

But no doubt, `middleman` is a great tool to build such a site. BTW: I used it for some internal projects lately which are a bit beyond websites: Building and Deploying [proxy.pac](http://de.wikipedia.org/wiki/Proxy_Auto-Config)-files + developing and building custom error-pages with inlined assets - javascript, images, stylesheets - for some some commercial proxy. I am very very satisfied that I chose `middleman`. @tdreyno Thanks for developing this: I really enjoy it using it on a daily basis.

[...]

PR #597, opened 2016-04-19
Attach stars count and pushed at for each GitHub repository

chaconnewu:
Added a separate file that is a complete clone of the original README.md file, adding the number of stars as well as last pushed date for each GitHub repositories included in the curated list. The date was collected through GitHub API (just now). And I can update the file regularly or as needed.

Let me know if you find it interesting. Thanks!

View the file directly at:
https://github.com/chaconnewu/awesome-ruby/blob/master/awesome_ruby_with_repo_info.md

### DIFF
--- a/awesome_ruby_with_repo_info.md
+++ b/awesome_ruby_with_repo_info.md
@@ -104,253 +104,253 @@ Thanks to all [contributors](https://github.com/markets/awesome-ruby/graphs/cont
 Abstraction
 -----------
 
--   [ActiveInteraction](https://github.com/orgsync/active_interaction) <span> | ★ 563, pushed 19 days ago | </span> - Manage application specific business logic.
--   [Apotomo](https://github.com/apotonick/apotomo) <span> | ★ 648, pushed 201 days ago | </span> - Based on Cells, Apotomo gives you widgets and encapsulation, bubbling events, AJAX page updates, rock-solid testing and more.
--   [Cells](https://github.com/apotonick/cells) <span> | ★ 2541, pushed 0 days ago | </span> - View Components for Rails.
--   [Decent Exposure](https://github.com/hashrocket/decent_exposure) <span> | ★ 1366, pushed 45 days ago | </span> - A helper for creating declarative interfaces in controllers.
--   [Interactor](https://github.com/collectiveidea/interactor) <span> | ★ 1340, pushed 25 days ago | </span> - Interactor provides a common interface for performing complex interactions in a single request.
+-   [ActiveInteraction](https://github.com/orgsync/active_interaction) <span> | ★ 563, pushed 20 days ago | </span> - Manage application specific business logic.
+-   [Apotomo](https://github.com/apotonick/apotomo) <span> | ★ 648, pushed 202 days ago | </span> - Based on Cells, Apotomo gives you widgets and encapsulation, bubbling events, AJAX page updates, rock-solid testing and more.
+-   [Cells](https://github.com/apotonick/cells) <span> | ★ 2546, pushed 1 days ago | </span> - View Components for Rails.
+-   [Decent Exposure](https://github.com/hashrocket/decent_exposure) <span> | ★ 1367, pushed 45 days ago | </span> - A helper for creating declarative interfaces in controllers.
+-   [Interactor](https://github.com/collectiveidea/interactor) <span> | ★ 1340, pushed 26 days ago | </span> - Interactor provides a common interface for performing complex interactions in a single request.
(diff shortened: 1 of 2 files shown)


markets:
hi @chaconnewu, first of all, thanks for that! I'm not really sure if this is the path we should take on this collection. I think it's interesting data (and even more metadata, as discussed here: #384, #304, #189), but I'm not really sure how to implement a good workflow to maintain it ...

It's probably a good fit for projects like:
* https://awesomelists.top/#/repos/markets/awesome-ruby
* http://ruby.libhunt.com/

Let me think about...

PS. do you used any open source script for that?

chaconnewu:
Hi @markets, thanks a lot for the feedback! I have three Python scripts to do this.

1. https://github.com/chaconnewu/dashboard_for_awesomes/blob/master/scripts/get_all_github_repos.py
It extracts all GitHub repository urls from your README.md.

2. https://github.com/chaconnewu/dashboard_for_awesomes/blob/master/scripts/get_repo_info.py
It pulls all repositories info using GitHub API and put it into MySQL database

3. https://github.com/chaconnewu/dashboard_for_awesomes/blob/master/scripts/attach_info_for_github_repos.py
It attaches stars count and last pushed date to the GitHub items, and generate a separete .md file.

Before our further discussion, I have to say that I'm an HCI researcher (https://scholar.google.com/citations?user=1TNPqLMAAAAJ&hl=en), currently conducting researches on curation repositories in GitHub. I'm doing this for research purposes, which is improving curation practice, and you are free to collaborate/respond to me or not.

I'm wondering why you are not sure about attaching the info in the list you maintained? It is a separate file, not overwriting your original README.md.

If you grant me write access to this repo, I can fully automate the process with one script. There are other options as well:
1. If you like, I can send out pull requests regularly (daily, weekly).
2. I can provide you a script, with which you can fully automate this.

What do you think?

elifoster:
I have two thoughts on this:
1. I think it is better for non-static content, such as the websites that @markets listed, as it does not require repeatedly updating the readme every day, even if it is automated with scripts. That can get seriously annoying for anyone watching this repository (over 300 people currently).
2. I think that this sort of brings attention away from what the library or project *does*, and directs it more to either A) how popular it is (stars), or B) how frequently it is pushed to (push date). The problem with (A) is the simple fact that a project's popularity has basically nothing to do with how useful it is. There are plenty of very useful gems that have only a few hundred downloads, and significantly less stars. The problem with (B) is that it might skew people's likelihood to look into a project if it hasn't been pushed to in a while, because often infrequent commits are associated with lack of maintenance, but it might be that the project is stable and does not depend on less stable, or more frequently updating, gems.

chaconnewu:
Thanks a lot for the details comments @elifoster ! 

For your first point, I'm wondering 1) what users are watching repositories for, and 2) whether users will really get annoyed for one activity trace update in their homepage. If they only watch for addition of content (I don't see how deletion of content can be of interesting), I can see it might be a problem.

For you second point, I agree with you, what users really want is quality, popularity, and project status. Unfortunately, GitHub does not provide all these features. So I can only provide some metrics for them to infer from. However, how the info is presented and how to design/calculate better metric is indeed crucial.

markets:
As I commented in other issues (https://github.com/markets/awesome-ruby/issues/304#issuecomment-72120807), quantitative metadata could be useful, but it's not all about numbers... I usually take, for example, downloads to get an idea about "use". 

On the other hand, periodically updates via PR doesn't seems to me like a great workflow 😅 , as @elifoster pointed out. Would be better to perform the updates somehow via a remote CI or something not manual... 

But after all, this is just a markdown document, probably the best idea is to build this apart, as a real application, something like http://libhunt.com. You should probably contact the team behind that project (https://www.libhunt.com/about).

markets:
A couple of related links more: 

* https://github.com/matteocrippa/awesome-swift/pull/530
* https://github.com/dkhamsing/awesome_bot

chaconnewu:
@markets thanks for the feedback and suggestions! Workflow seems like the concerns of every awesome-* owner I contacted. I've actually hosted awesome-* on a separate site a year and half before, but found out the GitHub context is much more important. Browsing/finding repositories on GitHub is what incepted in software developer's mind for a long time. I saw a lot of separate sites that use awesome-* as sources and nobody actually uses. 

Anyway I'll close this now and thank you very much for your time!

markets:
Thanks to you @chaconnewu for all those ideas and the work you have done! Keep it up! Do not hesitate to contact me if any related news happens regarding awesome-* collections, that was really a trending thing over the past 2 years.

2 more:

* https://zeef.com/
* https://www.g2crowd.com/

Curated collections everywhere!

Hope it helps.

PR #601, opened 2016-04-27
Add Rouge

Erol:
[Rouge](https://github.com/jneen/rouge) - A pure Ruby code highlighter that is compatible with Pygments.

RubyGem: https://rubygems.org/gems/rouge
Site: http://rouge.jneen.net


markets:
Nice one @Erol!

PR #608, opened 2016-05-13
Add monban

atbyrd:


markets:
Hi @atbyrd, as per contribution guides, 1 link per PR please.

atbyrd:
These two commits were about 6 months apart. I've remove the plugin as you suggested, makes sense.

markets:
Thanks @atbyrd. What are the pros of `monban` compared to other popular gems, like for example, `clearance`?

halogenandtoast:
> What are the pros of monban compared to other popular gems, like for example, clearance?

It's not an app engine. It's designed to be a authentication toolkit for building the authentication system you need (or you can use monban-generators to get a decent pre-implemented authentication system).

If you want to know more, I'd suggest checking out the README: https://github.com/halogenandtoast/monban

markets:
Thanks @halogenandtoast! 🚀

PR #611, opened 2016-05-22
Add Rodauth to Authentication

janko-m:
[Rodauth](https://github.com/jeremyevans/rodauth) is an authentication and account management framework for Rack applications. It's an alternative to Devise, Authlogic, Sorcery and others. Its advantages are:

* Full-featured, even has 2FA built-in
* Password hashes can't be accessed by the app database user (which means an attacker cannot access them even through SQL injection)
* Can be used with **any web framework** and **any ORM** (all other solutions are Rails-specific)
* All authentication logic can be encapsulated within a single Roda app
* Has support for JSON APIs, meaning all authentication & account management features are translated into the JSON API context (none of the other solutions have this ability)
* Extremely flexible


markets:
Hi @janko-m,

Thanks for sharing this resource! Really nice library 🚀  👍

markets:
PS. Really helpful PR description!

PR #613, opened 2016-06-05
Add dry-rb

raventid:
Motivation: dry-rb is a growing in popularity collection of libraries. A significant part of these libraries are well suited for working with ORM and persistence tools. But I found it a bit more fair to put them in Abstraction part of list as you can find very nice dependency injection containers among dry-rb libraries.

Also I've decided to add all of the libraries with one link to dry-rb organization as I think it will be a consistent approach according to my reflections(that these are toolkit for increasing abstraction level).


markets:
Thanks a lot @raventid! some of `dry-rb` gems are really amazing!

PR #614, opened 2016-06-09
Add braintree gem

jtdowney:
The braintree gem has been around for many years and has over 2 million downloads.

Disclosure: I work at Braintree but that doesn't stop it from being awesome ruby 😉


markets:
That's really true @jtdowney thanks!!

PR #616, opened 2016-06-16
add Ilovespec

wendy0402:
add Ilovespec, to make all your rspec test passed


markets:
Hi @wendy0402, thanks for sharing this! That's really funny stuff :) but we can't consider it as an essential tool to write Ruby apps.

wendy0402:
Hi @markets, no problem then, i just wany to share my gem, I'll close it and thanks for you to check it :)

PR #562, opened 2016-03-13
add new ruby learning resource

Khwilo:
This is a collection of various learning resources for Ruby.  GitHub repository for the site is at [iwanttolearnruby](https://github.com/awagener/iwanttolearnruby). Also include Practicing Ruby learning resource.

PR #620, opened 2016-07-14
Add Rubyfu book in References

KINGSABRI:


markets:
Hi @KINGSABRI, thanks for sharing this resource. Btw, as commented here https://github.com/markets/awesome-ruby/pull/562#issuecomment-198968411, I don't want to start a list of books and learning material.

PR #623, opened 2016-07-16
Added Shrine to file uploaders

paulanunda:
I came across Shrine today while doing research on ruby file uploaders. I think it deserves to be listed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -510,6 +510,7 @@ Thanks to all [contributors](https://github.com/markets/awesome-ruby/graphs/cont
 * [PaperClip](https://github.com/thoughtbot/paperclip) - Easy file attachment management for ActiveRecord.
 * [rack-secure-upload](https://github.com/dtaniwaki/rack-secure-upload) - Upload files securely.
 * [Refile](https://github.com/refile/refile) - A modern file upload library for Ruby applications, Refile is an attempt by CarrierWave's original author to fix the design mistakes and overengineering in CarrierWave.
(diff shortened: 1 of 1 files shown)


janko-m:
Just to briefly explain what makes Shrine special 😃 

Probably the biggest advantage of Shrine is its proper support for **background jobs**. While carrierwave_backgrounder doesn't work with direct S3 uploads, Shrine's backgrounding works for any combination of temporary and permanent storages. Also, Shrine also has *deleting* files in the background, which neither carrierwave_backgrounder nor delayed_paperclip support (which IMO is a big limitation if you're storing files on S3).

Shrine also has abstractions for **direct uploads**, so in addition to Amazon S3, you can upload to any other service which supports it, like Cloudinary, Uploadcare or Transloadit. Together with backgrounding you can give your users a fully asynchronous file attachment user experience.

Shrine is also the first file upload library with **logging** support. It logs uploading/processing/deleting separately, including the durations and the whole context (e.g. record class & ID). This really helps in monitoring performance, and it also makes debugging a whole lot easier.

There are countless more advantages of Shrine, which I intend to make a series of posts about, but I want to keep it brief here. In general Shrine was created because CarrierWave and Paperclip frankly aren't well designed, so important changes are very difficult to get in, extensions don't work with each other (since they're often just hacks),  and so on. Shrine was meant to be the next step, learning from mistakes of other libraries, and being something that will work for everyone.

markets:
Nice explanation @janko-m, thanks you guys!

PR #624, opened 2016-07-16
Add rack-time-zone-middleware into list

merqlove:


markets:
hi @merqlove, thanks for sharing! Actually, a bit fresh for this list!

merqlove:
@markets yeah, but first star achieved :) i did it because found this pattern really helpful for application performance. we can close this PR currently, i will reopen it on some progress.

markets:
> we can close this PR currently, i will reopen it on some progress.

100% agree

So, for example, in a Rails application, is this approach significantly performant rather than a controller around_action callback?

merqlove:
@markets, anyone able to write same logic in the controllers (via plain methods, concerns, services, ...), but because of cookies parsing, i found that really helpful to split this code into a middleware, because it is unrelated to the business logic. As we do it with Warden or Devise.

PR #635, opened 2016-09-23
Add Syro to Web Frameworks list

evanleck:
Syro is a "Simple router for web applications." that handles basic routing similarly to [Sinatra](https://github.com/sinatra/sinatra). It's by the [same author](https://github.com/soveran) as [Cuba](https://github.com/soveran/cuba) and according to him:

> An initial idea was to release a new version of Cuba that broke backward compatibility, but in the end my friends suggested to release this as a separate library. In the future, some ideas of this library could be included in Cuba as well.

[Source](https://github.com/soveran/syro#trivia).


markets:
Thanks @evanleck

PR #643, opened 2016-10-13
Add tty-prompt a CLI Builder to README

sabatesduran:


markets:
Thanks for sharing @sabatesduran! 

Since we already have TTY listed, I don't think we should add all (by the way, really amazing and independent) TTY plugins (https://github.com/piotrmurach/tty/blob/master/tty.gemspec#L20-L30). This list can grow a lot if we start adding framework's plugins...

sabatesduran:
You are right, didn't see that it was a dependency.
Thanks for the reply.

PR #644, opened 2016-10-14
Add RubyTest to Testing section

trans:


markets:
Hi @trans and thanks for sharing!

I'm not sure about this resource. Seems little active in last 2 years, and to be honest, it's probably too obscure for this collection...

trans:
It's the test framework used by Ruby Facets.

PR #647, opened 2016-10-27
Add HexaPDF to the PDF section

ramirio:

PR #648, opened 2016-10-27
Add Procodile to the Misc section

rmrgh:


markets:
Thanks @ramirio! Similar to #650 and #647

PR #649, opened 2016-10-27
Add EM Midori to the Web Frameworks section

rmrgh:


markets:
Thanks! Situation similar to #650

PR #654, opened 2016-11-11
Update clockwork

knoxknox:
Update link and description for gem 'clockwork'.
Looks like this gem is not maintained by original authors.

### DIFF
--- a/README.md
+++ b/README.md
@@ -869,7 +869,6 @@ Best suited for map-reduce or e.g. parallel downloads/uploads.
 
 ## Scheduling
 
-* [clockwork](https://github.com/adamwiggins/clockwork) - A scheduler process to replace cron.
 * [minicron](https://github.com/jamesrwhite/minicron) - A system to manage and monitor cron jobs.
 * [resque-scheduler](https://github.com/resque/resque-scheduler) - A light-weight job scheduling system built on top of Resque.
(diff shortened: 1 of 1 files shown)


markets:
hi @knoxknox, but the original source points out:

> This fork is no longer maintained, maybe try this one: https://github.com/Rykian/clockwork

But I also found this: https://github.com/Rykian/clockwork/issues/1.

So I'm not sure if this is really an option right now...

knoxknox:
Hi @markets, yes I saw this, but this is one of the authors of this gem.

What is your suggestion ?
Also I'm not sure about rubygems. Looks like version in rubygems also is not maintained.

markets:
Yes, there is actually some uncertainty... I also found this thread on Reddit: https://m.reddit.com/r/rails/comments/5a68on/what_happened_to_the_clockwork_gem 

So, probably right now best movement is to remove it from the list (we can add it back in case the project "revives" again). Btw, zhong seems a good lib for a distributed solution: https://github.com/nickelser/zhong

knoxknox:
@markets done, gem was removed from list

markets:
Thanks @knoxknox, this is a proper move IMHO!

There is also a discussion here, in the original repo: https://github.com/adamwiggins/clockwork/issues/23

PR #656, opened 2016-11-21
Add montrose

rossta:
https://github.com/rossta/montrose, a simple library for expressing, serializing, and enumerating recurring events in Ruby


markets:
Thanks @rossta nice one!!

rossta:
Thanks, @markets!

Issue #530, opened 2016-01-07
Rails app builder


kurenn:
Platter is a command line rails app generator to help you build Rails barebones as fast as possible. It includes devise and some other cool stuff. https://github.com/IcaliaLabs/platter

markets:
Hi @kurenn, we already have a pretty decent list of [Rails generators](https://github.com/markets/awesome-ruby#rails-application-generators). What are the advantages of `platter` regarding the others?

kurenn:
Well, is a simple setup we use at Icalia, it just adds a simple configuration for a testing environment, error handling, user handling, figaro for environment variables. It speeds things up at the beginning, this way every project we have follows the same bare bones.

markets:
Yes, apps generators are really useful, but current projects in that category seems more well-known and feature rich. Will close this one.

PR #659, opened 2016-11-22
Rails app generator

khusnetdinov:
Added one more rails generator


markets:
Thanks for sharing this @khusnetdinov!

A bit fresh for now, only 1k downloads according RubyGems. 

What are the pros against the other projects in that category? 

Similar situation to #530

khusnetdinov:
Each project has the same features with other. Gem differs from other by possibility create app using default flags, for example --api for creation little bit another application scaffold, manage predefined composition of Gemfile by using additional options (3 in TODO, and possible to add more). It also allow you to build on top Docker container with "12 factors rules" - passing and checking ENV variables. I think it's a more flexible and allow you to use Docker in "right way".

markets:
Thanks for your answer @khusnetdinov! Let's wait some time to see if the project reaches some popularity and adoption.

khusnetdinov:
Ok! Please don't close MR.

PR #660, opened 2016-11-23
add fix gem

cyril:


markets:
Nice project @cyril!

Taking into account the amount of testing related projects, what are the pros of this framework compared to the others and vastly popular/used frameworks?

cyril:
Glad to read you. Yes, sure:

Fix allows developers to write qualified specs with some [RFC 2119](https://www.ietf.org/rfc/rfc2119.txt)'s keywords:
* MUST
* SHOULD
* MAY

Unlike RSpec, Fix code is damn small (174 LOC), understandable, and trustable at the end of the day. And unlike minitest, Fix is written in a modular manner.

DRYed up spec and spec's description, Fix syntax looks like a logic-less document.

[Easily hackable](https://github.com/cyril/r_spec) (in action [here](https://asciinema.org/a/29172) or [there](https://asciinema.org/a/29070)), its output is colored by default.

There's also a quick presentation of this gem here: http://fixrb.org/

Issue #663, opened 2016-11-30
Don't know where should be repo?


khusnetdinov:
I collect information for ruby language with examples. It includes OOP, Patters, DS, Algorithms, Threads with references to source and books. 
Look here https://github.com/khusnetdinov/ruby.fundamental

Where can I put this?

fanjieqi:
Maybe the [resources](https://github.com/markets/awesome-ruby#resources).

khusnetdinov:
Thanks, I will send PR

khusnetdinov:
PR https://github.com/markets/awesome-ruby/pull/664

khusnetdinov:
Were problems with previous PR, fixed now only one commit.

markets:
Hi @khusnetdinov, thanks for sharing! I think it would fit better under [Coding Style Guides](https://github.com/markets/awesome-ruby#coding-style-guides), as it's similar to https://github.com/franzejr/best-ruby. Also, in resources we add different stuff:

> Where to discover new Ruby libraries, projects and trends.

If you guys are agree, we just need to update #666 (😈 )

khusnetdinov:
Updated!

khusnetdinov:
Ping, Are guys agree or not?

markets:
Merged @khusnetdinov 👍

PR #666, opened 2016-12-01
fundamental

khusnetdinov:
Fix and PR relate https://github.com/markets/awesome-ruby/issues/663


markets:
Thanks @khusnetdinov!

PR #667, opened 2016-12-11
Add jaro_winkler link

tonytonyjan:
Some remarkable points:

- The only one jaro_winkler gem with C extension that can handle UTF-8 string so far.
- Part of [did_you_mean](https://github.com/yuki24/did_you_mean/blob/62a666ccbc21ded6b471ba883b935050b890f809/Gemfile), which has become built-gem since Ruby 2.3.
- Part of [roma](https://github.com/roma/roma/blob/a0311401d4e3fe17e218523c0e154084b55ca2e8/roma.gemspec#L48)


markets:
Thanks @tonytonyjan!

PR #673, opened 2016-12-29
Update Sorcery to new org repo and added Rails 5

Ch4s3:
We moved sorcery over to an org repo, and awesome ruby should reflect its new home.


markets:
Thanks @Ch4s3 👍

PR #676, opened 2017-01-04
Add Deveo

Markowy:


markets:
The website redirects me to another domain... Seems a similar service, but to be honest, I'm not sure how specific is this for the Ruby community. I'm not saying this is not useful at all, but I'd keep this "Services" category small, with only most relevant services, otherwise this section could grow too much.

PR #677, opened 2017-01-05
Add terjira gem

keepcosmos:
Add Jira command-line power tool.


markets:
Nice tool @keepcosmos! thanks for sharing 🚀

PR #680, opened 2017-01-12
remove upmin-admin-ruby: framework is no longer actively maintained

Fivell:
Maintainers  suggest to use Administrate


markets:
Thanks @Fivell

PR #684, opened 2017-01-21
add simple csv creator

wendy0402:


markets:
Hi @wendy0402! thanks for sharing :+1:

Still to fresh for this list, only 130 downloads according to RubyGems, so let's keep it open some time and see how it evolves.

markets:
Closing for now @wendy0402 (still under 1k downloads after one year), please re-open in case you want to resume that discussion (or the project reaches higher popularity/use). Thanks!

PR #688, opened 2017-02-10
Add tabulous

wynksaiddestroy:


scudelletti:
The maintainer recommends to not use it anymore, and probably it won't support new major versions of rails.

https://github.com/techiferous/tabulous/blob/master/FUTURE.md

wynksaiddestroy:
At least all versions of Rails 5 are going to be supported. If you look at the other gems in the navigation category, you'll see that the situation is much worse there:
The last commit of [active_link_to](https://github.com/comfy/active_link_to) happened over two years ago. Its documentation is still speaking of Rails 3.
[simple-navigation](https://github.com/codeplant/simple-navigation) was also updated two years ago.
In comparison to these competitors tabulous is the most actively developed gem in the (Rails) navigation category.

markets:
Hi @scudelletti @wynksaiddestroy

Both, active_link_to and simple-navigation were active in the last few days, even they released new versions.

PR #689, opened 2017-02-11
Add rails_or

khiav223577:
RailsOr helps those who want to do an OR query in Rails 3+ ActiveRecord, 
which is only available in Rails 5. ( https://github.com/rails/rails/pull/16052 )


markets:
Hi @khiav223577, thanks for sharing your project to this list!

Btw, I think we can't consider this gem an essential/top resource to build Ruby apps.

khiav223577:
Okay, I didn't notice that the list required `essential` ruby library.

markets:
sorry @khiav223577, not just "essential" libs (is really hard to draw a line here), but in that case is more syntax sugar for old AR versions... if we start to add all kind of gems, the list can grow a lot and lose value IMHO. Also, it has only 1k downloads according RubyGems.

PR #691, opened 2017-02-15
add slideshow (s9) gem in presentations section

geraldb:
Thanks for the awesome list. I would like to add the slideshow (s9) gem that lets you generate presentations / talks / slideshows from your text sources (in markdown) with web templates. Cheers.


markets:
Nice one @geraldb! Thanks 🚀

PR #693, opened 2017-02-23
Add state-machines to the state-machines list

cefigueiredo:
Add also its ActiveRecord integration plugin.


markets:
Thanks for contributing @cefigueiredo! We don't need to add specific adapters, could you please review it?

PR #695, opened 2017-03-16
Add milia

jekuno:
Adds milia, a non-invasive multi-tenanting gem for Rails which supports Devise authentication out of the box.


markets:
Thanks @jekuno 🚀

PR #698, opened 2017-03-16
Added Nanobox as a DevOps tool

sdomino:


markets:
Hi @sdomino, thanks for your contribution!

I think it fits better under Service and Apps, do you agree? I realized that point comparing to other projects under Devops and Environment categories.

PS Actually the relation between resource <=> category is one-to-one

sdomino:
Ahh, you're right. It does make much more sense under Services and Apps. Thanks!

PR #705, opened 2017-04-10
Add Module Linker.

fiatjaf:
Adds [Module Linker](http://fiatjaf.alhur.es/module-linker/#/ruby), a Chrome Extension that adds direct links to `require` statements right in GitHub source code. What the extension actually does is exemplified better in the following animated GIF:

### DIFF
--- a/README.md
+++ b/README.md
@@ -1156,6 +1156,7 @@ Online tools, services and APIs to simplify development.
 * [HoundCI](https://houndci.com) - Review your Ruby code for style guide violations.
 * [HuBoard](https://huboard.com) - Kanban board for GitHub issues.
 * [Inch CI](http://inch-ci.org/) - Documentation badges for Ruby projects.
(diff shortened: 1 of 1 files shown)


markets:
Thanks for sharing it @fiatjaf!!

I'm actually using https://github.com/OctoLinker/browser-extension,it works on ff and opera too and supports other statements like `gem`.

fiatjaf:
Ok, since you're saying that, I let me advertise my product better and distinguish it from the competition, then :stuck_out_tongue: 

Module Linker also supports `gem` (see screenshots on the page), and Firefox support is coming. The remarkable feature it has that Octolinker doesn't is that it adds direct "normal" `<a href=...>` links, while Octolinker calls an HTTP endpoint at the time you click, so you can't see where the click will take you and middle clicks don't work.

Also, I've used Octolinker for a while and generally saw much more errors than Module Linker has shown me, in terms of not finding the correct URL to link to, talking about Ruby specifically. These are very rare nowadays in Module Linker.

markets:
Thanks @fiatjaf! Let's leave it open for some time to let community leave their opinion (and to wait Firefox support).

PR #711, opened 2017-05-27
Add feedparser - A feed parser and normalizer (Atom, RSS, JSON, etc) library.

geraldb:
Hello,
   Thanks for this awesome ruby list. I'd like to add the feedparser gem (@ [github.com/feedparser/feedparser](https://github.com/feedparser/feedparser)) that just got updated to include the newest JSON Feed format (in addition to the "oldies" RSS and Atom). The library gets used by the pluto feed reader (among others) and runs, for example, the OpenStreetMap Blogs. Again thanks for the awesome ruby list. Keep it up. Cheers.


markets:
Thanks to you @geraldb! You're also maintaining pretty useful collections and projects 🔝

PR #716, opened 2017-07-18
Add Shopr e-commerce CMS gem (Rails 5, Devise)

kirylpl:
Add Shopr e-commerce CMS gem.


kirylpl:
@markets  Please check it to approve

markets:
Hi @kirylpl, thanks for contributing! 

Actually this fork has very few downloads (just 1k downloads according to RubyGems) and has little activity (no commits) during last year. So right now, we can't consider it an essential tool for the Ruby ecosystem. Let's wait some time and we'll see if the project finally takes up.

kirylpl:
@markets Look, original Shoppe gem got over 35k+ downloads on Rubygems, but now it's abandoned.
This fork solve two needs - Rails 5 and Devise auth support.

If gem will not be known even through 'awesome list', how it will became popular?
Please let people to chose on their own, this gem is pretty fair and modern, but only monstrous Shopify and pretty old Chameleon is on hearing.

markets:
Hi @kirylpl, but imagine if we start to add all new projects to these collections... this is not the idea of this "awesome list" trend actually.

Some other comments:
- site returns me a DNS error https://shoprgem.com/ 
- it's not clear how I should find/browse docs about the project

Issue #719, opened 2017-07-29
Wrong isnt maintained any more


Kapenaar:
It seems that the Wrong test assertion project is no longer maintained. The last commit was in 2014, there are long list of critical open issues and even the hello world example doesn't work with the latest release [see here](https://github.com/sconover/wrong/issues/46).

I think this should be removed from awesome-ruby as it is not awesome any more, just frustrating.

markets:
@Kapenaar agreed! make sense to remove it from here. Could you please send a PR? Thanks!

PR #721, opened 2017-08-01
Removed "Wrong" as it is no longer maintained (#719)

Kapenaar:
As proposed in ticket #719, the "Wrong" project is no longer maintained, and should be removed.


markets:
Thanks @Kapenaar

PR #724, opened 2017-08-08
Add batch-loader

exAspArk:
Hi!

I created a new gem called [batch-loader](https://github.com/exaspark/batch-loader) which helps to avoid N+1 DB queries, HTTP requests, etc. In addition to the detailed description and some ASCII art in the [README](https://github.com/exaspark/batch-loader) I also created a small presentation. I'm using the [slides](https://speakerdeck.com/exaspark/batching-the-powerful-way-to-solve-n-plus-1-queries-every-rubyist-should-know) to describe to my internal team why the gem was created and how we can use it.

The gem can be used with any Ruby code, not only with GraphQL. In fact, we're using it with RESTful APIs on production. So, I was thinking about adding it to another category, the closest one could be `Data Processing and ETL`. But I decided to leave it in `GraphQL` for now, since it's not about ETL.

Would be glad to get your feedback. Whether it should be in `GraphQL` (it's fine for me), in some other existing categories or it's better to create a new one.

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,7 +76,7 @@ Thanks to all [contributors](https://github.com/markets/awesome-ruby/graphs/cont
   * [Presentation Programs](#presentation-programs)
   * [Process Monitoring](#process-monitoring)
   * [Processes and Threads](#processes-and-threads)
-  * [Profiler](#profiler)
+  * [Profiler and Optimization](#profiler-and-optimization)
   * [Queue](#queue)
   * [Rails Application Generators](#rails-application-generators)
   * [Robotics](#robotics)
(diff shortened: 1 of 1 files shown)


exAspArk:
Hi, I also published a post about [batching for Rubyists](https://engineering.universe.com/batching-a-powerful-way-to-solve-n-1-queries-every-rubyist-should-know-24e20c6e7b94) :)

markets:
Thanks for sharing this @exAspArk!

I'm not sure about section... I'd like to see this outside of GraphQL category.

Bullet is listed under [https://github.com/markets/awesome-ruby#profiler](Profiler), do you think make sense to move it there?

exAspArk:
Hi @markets!

Thanks for your feedback!

Alternatively, I think that [batch-loader](https://github.com/exaspark/batch-loader) can be moved to the category like "Performance Optimization" (just pushed the changes to demonstrate how it can look like). In the future, we can add some other tools in the same category. For example, [bootsnap](https://github.com/Shopify/bootsnap) – a gem which speedups Rails boot time with some optimizations and caching.

Please let me know what do you think about the idea of the new category?

markets:
Hi @exAspArk, 

I just thought a new idea, I'd go with renaming `Profiler` to `Profiler and Optimization` and add it there. What do you think? It make sense to me at this point, we can split it in the future if the section grows a lot. Thanks.

exAspArk:
@markets great idea! We can always split the sections in the future if it's necessary. Pushed the changes.

PR #725, opened 2017-08-09
added ActiveEndpoint to Performance monitoring

khusnetdinov:
ActiveEndpoint gem allow you to track request and responses for you rails application. It keep all information and all params in each probe that stores in db, its works with minimal affecting application. If application was requested for route that was not in routes it also keep. Can simply define black list and other settings for keeping all clear. Use full for start optimize you application for response time.

### DIFF
--- a/README.md
+++ b/README.md
@@ -827,6 +827,9 @@ Thanks to all [contributors](https://github.com/markets/awesome-ruby/graphs/cont
 
 ## Performance Monitoring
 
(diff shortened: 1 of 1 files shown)


markets:
Hi @khusnetdinov, thanks for contributing 👍 

Still to fresh for now:

> Attention! Gem in under active test and is preparing for first release !

PR #740, opened 2017-09-21
Add Docile

ms-ati:

## Project

[Docile](https://github.com/ms-ati/docile)

## What is this Ruby project?

A tiny library that lets you map a DSL (domain specific language) to your Ruby objects in a snap.

I'm not aware of any other libraries that fill the same niche as Docile. It was written to extract the common patterns around usage of chained contexts for `instance_exec` or `instance_eval` from ad-hoc DSL code in various libraries. The goal is to have a common, community-maintained implementation of these patterns that avoids the common errors observed in the wild, such as local variables from the host object being unavailable to the code in the given block.

This library is fairly widely used as a dependency for other projects to implement their own configuration DSLs, with [SimpleCov](https://rubygems.org/gems/simplecov) probably being the most notable.

---


markets:
Cool project @ms-ati! Thanks!

PR #747, opened 2017-09-26
Add MicroMachine to "State Machines" section

evanleck:
## Project

[MicroMachine](https://github.com/soveran/micromachine)

## What is this Ruby project?

A small, simple finite state machine implementation.

I think the README lays it out pretty simply:

> There are many finite state machine implementations for Ruby, and they all provide a nice DSL for declaring events, exceptions, callbacks, and all kinds of niceties in general.
>
> But if all you want is a finite state machine, look no further: this has less than 50 lines of code and provides everything a finite state machine must have, and nothing more.


markets:
amazing library, again! @soveran 👌

PR #678, opened 2017-01-05
Added stemming library

abdullahtariq1171:
Stemming libraries are used in pre-processing of data.

PR #750, opened 2017-09-29
Add Bitwise

kenn:
## Project

[Bitwise](https://github.com/kenn/bitwise)

## What is this Ruby project?

Fast, memory efficient bitwise operations on large binary strings.

Does not crash :) Faster, stabler, and more memory efficient.

---


markets:
Hi @kenn, in this case, I think this is probably too obscure for this list (see examples #289 #678) and there is really little activity in the latest years (last release was on 2013). So let's keep open for some time and let people comment, but tbh I'm not sure if this fits well in that list.

kenn:
Yeah I agree that not everyone needs it, but no activity just means it's incredibly made stable (unlike other libraries) and the API is mature enough. I would consider bit vectors are essential data type.

One of the common use cases is an algorithmic data structure for something like [unique popcount](http://blog.getspool.com/2011/11/29/fast-easy-realtime-metrics-using-redis-bitmaps/) if you do the same thing with RDB, it would be prohibitively expensive - 1M users would only need 125KB compared to 16MB per day.

There are many other use cases, like bloom filters and other machine learning algos, which is the blind spot of Ruby compared to Python.

PR #752, opened 2017-09-29
Add Sunzi

kenn:
## Project

[Sunzi](https://github.com/kenn/sunzi)

## What is this Ruby project?

Server provisioning utility for minimalists

All you need is plain shell scripts. No clunky Ruby DSLs or YAML scripting compared to Chef, Puppet, Ansible or Salt.

---


markets:
Interesting project, but I think most of people is using the projects you mentioned: Chef, Puppet, Ansible, ... 

Itamae (#732) seems also a lightweight version of Chef, but with more downloads (usage), community behind (11 vs 50 contributors), and recent activity/releases.

kenn:
Yeah. Those popular tools are designed for large projects, while Sunzi is designed for small or solo projects. Itamae is also from a big, public company. Even for solo developers, it's a good habit to not directly work on servers for admin stuff, even when only you have the access.

Sunzi has the lowest learning curve as there's really nothing to learn, like any cookbooks or DSL. It's just shell scripts - as you find them with Google, just copy-paste them without converting to DSL.

PR #756, opened 2017-09-29
Add EnumAccessor

kenn:
## Project

[EnumAccessor](https://github.com/kenn/enum_accessor)

## What is this Ruby project?

Simple enum fields for ActiveRecord.

Rails 4.1's official implementation came out with a design flaw. This gem does not have such crufts, and also supports a lot more features like validation, where-scoping, translation and forms with even shorter code base.

---


markets:
Hi again @kenn, IMO and as you said, this is just some extra (and cool!) stuff compared to the official Rails `enums`, so at first I'm -1 on adding these kinds of gems to that list, we can't consider it essential Ruby. 

Thanks for yours shares by the way!

kenn:
Got it, understood. Thanks for your comment!

PR #759, opened 2017-10-02
Add connection_pool

matheusteixeira:
## Project

[connection_pool](https://github.com/mperham/connection_pool)

## What is this Ruby project?

This project makes it easy to create connection pools for any database. ActiveRecord has it own Connection Pool, but you can use this project to create pooling for Redis, Dali or any Ruby network client.

- This is a very well tested, well estabilished and with lots of contribuitors
- It's simple to use
- It's maintained by [Mike Perham](https://github.com/mperham), the author and maintener of **Sidekiq**.

---

### DIFF
--- a/README.md
+++ b/README.md
@@ -383,6 +383,7 @@ Thanks to all [contributors](https://github.com/markets/awesome-ruby/graphs/cont
 
 ## Database Tools
 
+* [connection_pool](https://github.com/mperham/connection_pool) - Generic connection pooling for Ruby, that can be used with anything, e.g. Redis, Dalli, etc.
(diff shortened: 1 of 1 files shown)


matheusteixeira:
Hi @markets, I noticed that you added a Label `add link`. Do I need to add anything else in this PR?

markets:
Hi, not really, I usually leave PRs open some time to let people vote, comment (doesnt happen much tbh) ... and also to take look to the project. 

But in this case, I also think this is 🔝 Ruby, so time to 🚀 

Thanks @matheusteixeira @mperham

PR #769, opened 2017-10-18
Add Fastsheet gem

dmk:
## Project

**Fastsheet** - https://github.com/dkkoval/fastsheet

## What is this Ruby project?

Fastest spreadsheet reader for ruby ([benchmark](https://github.com/yivo/ruby-excel-readers-benchmarks/blob/master/index.rb)).

* **speed:** reading **10 000**-rows file was approximately **40 times faster** than any similar gem.
* **minimal memory usage:**
    * the sheet is read only once
    * the sheet's instance stores only the initial array of cells read

---


markets:
Thanks for sharing @dkkoval 👍  Still too fresh for this list (900 downloads), but it can be a powerful library if you need to deal with big spreadsheets. The Rust dependency isn't desirable, could make it harder to deploy it anywhere... but Rust is growing up fast lately, so ¯\_(ツ)_/¯ Let's see what community thinks.

PR #770, opened 2017-10-23
add Neovim gem

GustavoSantosFe:
this an awesome gem that helped me to make some plugins to make coding easier it's pretty easy to use and it's one of the few reasons to move from vim to neovim

## Project

Neovim - https://github.com/alexgenco/neovim-ruby

## What is this Ruby project?

ruby wrapper for neovim to make plugins for the editor in ruby

don't know any similar ones

---


markets:
Thanks for sharing @GuSilverFlame! This is surely a very powerful tool for Neovim users (seems well maintained project too), but to be honest, I'm a Sublime user (ok, Vim occasionally, in servers) so I have no idea how essential is this to the Ruby ecosystem. Let's see what community thinks.

PR #785, opened 2017-12-14
Add Sinja

jnylen:
## Project

Sinja
https://github.com/mwpastore/sinja

## What is this Ruby project?

It builds an JSONAPI api using jsonapi-serializers and controllers.

It's built for Sinatra rather than Rails/Hanami.


markets:
Hi @jnylen and thanks for your contribution 👍 

This is a nice Sinatra "extension", but I'm not sure if we can consider it essential Ruby or a very used resource (~6k downloads). Let's keep it open and leave community drop their comments!

PR #789, opened 2017-12-22
Add Rails Event Store

fidel:
## Project

Rails Event Store a.k.a. RES - https://github.com/RailsEventStore/rails_event_store
 
## What is this Ruby project?

Rails Event Store (RES) is a library for publishing, consuming, storing and retrieving events. It's your best companion for going with an event-driven architecture for your Rails application.

I think that none of similar tools are present on the list currently. I proposed a separate category, but it can probably fit into `Abstraction` one. 

You can use it:

* as your Publish-Subscribe bus
* to decouple core business logic from external concerns in Hexagonal style architectures
* as an alternative to ActiveRecord callbacks and Observers
* as a communication layer between loosely coupled components
* to react to published events synchronously or asynchronously
* to extract side-effects (notifications, metrics etc) from your controllers and services into event handlers
* to build an audit-log
* to create read-models
* to implement event-sourcing

---

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,6 +55,7 @@ Thanks to all [contributors](https://github.com/markets/awesome-ruby/graphs/cont
   * [Encryption](#encryption)
   * [Environment Management](#environment-management)
   * [Error Handling](#error-handling)
+  * [Event-driven architecture](#event-driven-architecture)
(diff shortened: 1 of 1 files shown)


markets:
Hi @fidel and thanks for your contribution! Cool project :+1: I'd like to move it under Abstraction for now, this new category seems too much specific to me.

PR #796, opened 2018-01-25
Added RoRvsWild

antoinem:
## Project

RoRvsWild
https://github.com/BaseSecrete/rorvswild

## What is this Ruby project?

RoRvsWild is a Ruby gem to monitor performances and quality in Rails applications.

This gem has a double mode: development, and production. 
It can be used without an account to profile your requests performances in your development environment.
It can also be used in your production and staging environments with an account on https://rorvswild.com. With such an account you also get extra benefits such as 30-day trace, background jobs monitoring, errors monitoring and notifications.

---


markets:
Thanks @antoinem 👍 nice project! In the past, I used to use new_relic for dev env, but at some point they removed this mode, so I think this is useful.

antoinem:
sorry I didn't notice the conflict. Fixed.

PR #797, opened 2018-01-31
Add oauth_token_verifier

shkrt:
## Project

oauth_token_verifier
https://github.com/Shkrt/oauth_token_verifier

## What is this Ruby project?

Simple abstraction library for validating OAuth access tokens obtained from third parties.

This library's only responsibility is validation of access token and returning some additional infos about its owner from OAuth provider. Useful for mobile application development, i.e. when OAuth flow starts from mobile app, and then somehow continues on the server side. The most obvious use case - mobile app user taps "Login via facebook" button in the app interface, then we get an access token from facebook, and the token goes to the server application, where we can verifiy the user identity and token validity, map user info fields to the ruby objects with the help of this library.
 
---


markets:
Thanks for sharing @Shkrt! Still too fresh for this list (~1k downloads, 4 stars). Let's keep it open for some time!

PR #798, opened 2018-02-09
Add Komponent

florentferry:
## Project

Komponent
https://github.com/komposable/komponent

## What is this Ruby project?

An opinionated way of organizing front-end code in Ruby on Rails, based on components

We strongly emphasis to made a tool simple to use, following the Rails way, to organizing views based on components.  We provide generators to install, and generate new components with some options to use template engine and stylesheet engine of your choice. 

---


markets:
Thanks for sharing @florentferry! Cool project! A bit fresh for now (only 1k downloads according RubyGems) for this list, but promising 👍  Let's keep open for some time to let people leave their feedback!

florentferry:
Seems good to me. Thanks for the review.

PR #799, opened 2018-02-09
Add ComfyBootstrapForm (Form Builder)

GBH:
## Project

ComfyBootstrapForm https://github.com/comfy/comfy-bootstrap-form

## What is this Ruby project?

It's a form builder that extends default `ActionView::Helpers::FormBuilder` in order to generate Bootstrap 4 markup for form elements.

Currently, this is the only form builder that properly supports Bootstrap 4 markup and works in Rails 5.2 by wrapping`form_with` form helper. Rails 5.0+ versions are also supported as `form_for` is wrapped as well. Other form builders are either very out-of-date or not planning to support Bootstrap 4 at all.

Now it's a direct dependency of [ComfortableMexicanSofa CMS](https://github.com/comfy/comfortable-mexican-sofa) that uses Bootstrap 4 in the admin area.

Cheers!

### DIFF
--- a/README.md
+++ b/README.md
@@ -593,6 +593,7 @@ Best suited for map-reduce or e.g. parallel downloads/uploads.
 
 * [Abracadabra](https://github.com/TrevorHinesley/abracadabra) - The gem that swaps out text with a fully-compliant Rails form in one click.
 * [Cocoon](https://github.com/nathanvda/cocoon) - Dynamic nested forms using jQuery made easy; works with formtastic, simple_form or default forms.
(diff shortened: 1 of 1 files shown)


markets:
Hi @GBH and thanks for sharing this! A bit fresh for now (only 1,7k downloads), let's keep open for some time to let the community leave their thoughts about it.

GBH:
Heya @markets, how about now? It's getting up there with downloads (84K) https://rubygems.org/gems/comfy_bootstrap_form

I think it's a pretty good alternative to another bootstrap form builder.

Thanks!

PR #810, opened 2018-05-03
Add ResponseCode

torokmark:
## Project

[ResponseCode](https://github.com/torokmark/response_code)

## What is this Ruby project?

ResponseCode is a simple lib that helps you handle response codes in a readable way instead of numbers.

Very small and handy without having dependencies.

---


markets:
Hi @torokmark and thanks for sharing! Still too fresh for this list. Let's keep it open for some time as we usually do.

PR #815, opened 2018-06-14
Add RIUI

lukecrum:
## Project
RIUI
https://github.com/lukecrum/RIUI
## What is this Ruby project?
Simple GUI Library
## Describe features.
RIUI makes it simple to create user interfaces with Ruby
It is much easier to use and has less dependencies
## Enumerate comparisons.
something like qtbindings is more capable, however has a much steeper learning curve. 


lukecrum:
on second thought, this is too new to put on here. might reopen on a later date.

PR #818, opened 2018-06-24
Add Resultr library on Abstration category

leocamelo:
## Project

https://github.com/leocamelo/resultr

## What is this Ruby project?

Resultr provides a simple interface to work with computation results, highly inspired by [Rust results](https://doc.rust-lang.org/std/result) for handling errors on function returns.

The main idea here is that a function unsuccessful result shouldn't be an exception, buts an error return.

---

### DIFF
--- a/README.md
+++ b/README.md
@@ -128,6 +128,7 @@ Thanks to all [contributors](https://github.com/markets/awesome-ruby/graphs/cont
 * [Light Service](https://github.com/adomokos/light-service) - Series of Actions with an emphasis on simplicity.
 * [Mutations](https://github.com/cypriss/mutations) - Compose your business logic into commands that sanitize and validate input.
 * [Responders](https://github.com/plataformatec/responders) -  A set of Rails responders to dry up your application.
+* [Resultr](https://github.com/leocamelo/resultr) - A simple interface to work with methods results.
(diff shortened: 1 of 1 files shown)


markets:
Hi @leocamelo and thanks a lot for sharing! Still too fresh for this list (only ~500 downloads according RubyGems). Let's keep it open for some, will see if the project reaches some popularity/use! 

Simple but nice api btw 👍 I was just reading about https://github.com/dry-rb/dry-monads

leocamelo:
Hi @markets , thanks for the feedback! I will read the dry-monads and keep improving the library. It helps me with pluga.co codes and maybe reaches some popularity. :)

jdickey:
@leocamelo Thanks for sharing this! I've used the gamut of interactor- and monad-style result objects in Ruby over the last 6-7 years, and this looks nice and simple. Could definitely see it coming in handy for some lightweight work.

@markets I've been using the dry-rb ecosystem for quite a while now. [dry-monads](https://github.com/dry-rb/dry-monads) are great, particularly in combination with other Gems in their stack. I see this coming in handy for something where _all you want_ is a success/failure indicator with a single (object) value giving further details. "Raw" dry-monads really want you to think like a Haskellite, and most Rubyists I've known really aren't. 😀

PR #823, opened 2018-07-25
Add Acra, database protection suite

vixentael:
## Project

Website: https://www.cossacklabs.com/acra/
Github: https://github.com/cossacklabs/acra/
AcraWriter ruby gem: https://rubygems.org/gems/acrawriter/

## What is this Ruby project?

Acra is a network encryption proxy to protect databases and database-based applications from data leaks, allowing greater security for distributed applications via cryptography and intrusion detection.

[AcraWriter](https://github.com/cossacklabs/acra/wiki/Using-Acra-to-Protect-Your-Rails-App) is a part of Acra suite, a client-side library, which integrates into the app flow either through ORM or directly, and provides the means to encrypt the sensitive data. AcraWriter is written for different platforms on different languages, including Ruby.

Acra doesn't have similar projects, however you might think about Vault's Encryption as a Service engine, but deployed in your own infrastructure, without leaking keys anywhere.

- Acra allows to selectively encrypt sensitive records with strong multi-layer cryptography in client-side application. Client-side app doesn't have _decryption_ key, thus compromising app won't lead to stealing key and decrypting data in the database.
- Sensitive data is encrypted before being transmitted to the database, and decryption key is not stored on client side.
- Data is decrypted in trusted environment (on AcraServer).
- Acra allows to detect potential intrusions and SQL injections (using SQL firewall and poison records).
- Has SIEM integrations.
- Suits for distributed, microservice-rich environments.
- Has numerous examples and tutorials for Ruby client applications.
- Allows your application to cover GDPR articles 25, 32, 33, 34. 
- Tested, supported, audited.

---

### DIFF
--- a/README.md
+++ b/README.md
@@ -425,6 +425,7 @@ Best suited for map-reduce or e.g. parallel downloads/uploads.
 * [Seed Fu](https://github.com/mbleigh/seed-fu) - Advanced seed data handling for Rails.
 * [Upsert](https://github.com/seamusabshere/upsert) - Upsert on MySQL, PostgreSQL, and SQLite3. Transparently creates functions (UDF) for MySQL and PostgreSQL; on SQLite3, uses INSERT OR IGNORE.
 * [SecondBase](https://github.com/customink/secondbase) - Seamless second database integration for Rails. SecondBase provides support for Rails to manage dual databases by extending ActiveRecord tasks that create, migrate, and test your application.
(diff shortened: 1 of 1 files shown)


markets:
I think we should close this one, after 1.5 years the gem has only 1k downloads, so it doesn't seem the Ruby ecosystem picked up this plugin.

PR #826, opened 2018-08-15
add flashee project

jbernsie:
## Project

Flashee
- https://github.com/jbernsie/flashee
- https://rubygems.org/gems/flashee

## What is this Ruby project?

Flashee provides a helper class and an ERB partial to render Rails flash messages with CSS classes dependant on the type of flash message being rendered. _i.e. when a error message is rendered attach the `danger` class._

Enumerate comparisons.

---


markets:
Hi @jbernsie and thanks a lot for sharing. Unfortunately, I think we can't consider this gem essential Ruby or a very commonly used resource (only 1k downloads according RubyGems), so not sure if that fits well in this collection...

jbernsie:
Fair enough man, no stress.

PR #834, opened 2018-09-05
Remove paperclip as it has been officially deprecated

drewpterry:
Paperclip is now deprecated and should not be used so it seems like it should be removed.

ref: https://robots.thoughtbot.com/closing-the-trombone

## Project

https://github.com/thoughtbot/paperclip

---

### DIFF
--- a/README.md
+++ b/README.md
@@ -584,7 +584,7 @@ Best suited for map-reduce or e.g. parallel downloads/uploads.
 * [attache](https://github.com/choonkeat/attache) - Standalone image and file server to decouple your app from file management concerns.
 * [CarrierWave](https://github.com/carrierwaveuploader/carrierwave) - Classier solution for file uploads for Rails, Sinatra and other Ruby web frameworks.
 * [DragonFly](https://github.com/markevans/dragonfly) - A Ruby gem for on-the-fly processing - suitable for image uploading in Rails, Sinatra and much more!.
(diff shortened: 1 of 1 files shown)


markets:
Hi @drewpterry and thanks a lot for contributing! 

I have doubts about this removal, at least at this point. This is a really big and used project (> 21m downloads) and it will be around for some time more (it wouldn't be easier to migrate huge apps). Similar situation as other important projects (like DataMapper) in the ecosystem: we still have it in the collection but with a comment/notice attached. Paperclip will still be relevant in the community for some time. Maybe instead of deleting it, we can add a comment for now, thoughts?

drewpterry:
Thanks for the quick reply @markets! What you says make sense so I am okay with simply adding it as a comment so I have changed it accordingly.

My thinking was that most people using this Repo want to see what the current Ruby ecosystem is like or are looking for new libraries to try out for their specific problems. So it doesn't seem that helpful to include a repo that is no longer maintained (since the only people it should be relevant to are the ones that are still currently using it).

Just my thoughts, but feel free to merge it as is with the deprecation notice!

markets:
Your thinking was totally fine and that's the main goal and idea behind this compilation! But in cases like this one, a project with millions of downloads, and recent releases to offer kind of latest "stable" release... I prefer to keep it in the list for some time more. Exceptions they said :smiley: 

Personally, I like CarrierWave, I fount it pretty stable, feature rich (lots of plugins too) and flexible enough for my current needs. Active Storage and Shrine are also very cool options :ok_hand: with really different approaches.

PR #838, opened 2018-09-13
Replaced GlobalPhone with telephone_number

kibenimatik:
Replaced [GlobalPhone](https://github.com/sstephenson/global_phone) with [telephone_number](https://github.com/mobi/telephone_number) since it is pretty outdated and has several issues.


markets:
Could you please add the new link in alphabeticals order as per contribution guides? Thanks

markets:
What about https://github.com/daddyz/phonelib?

PR #849, opened 2018-10-06
Removed Green Ruby News from resources section.

bkuhlmann:
## Overview

Unfortunately, this service is no longer running and can be removed
from the list of choices. The feed was ended on
[2017-06-04](http://greenruby.org).


markets:
Thanks @bkuhlmann!

PR #854, opened 2018-10-21
Add new Optimizations section with some awesome gems :)

n-rodriguez:


n-rodriguez:
I've just seen there's already a "Profiler and Optimization" section... This PR might be redundant... But it's not exactly the same purpose.

markets:
Thanks! Yes, that section has a totally different purpose.

n-rodriguez:
Thank you!

PR #904, opened 2019-03-08
Add discard gem

lautarocastillo:
## Discard gem

https://github.com/jhawthorn/discard - Soft deletes for ActiveRecord done right.

## What is this Ruby project?

A simple ActiveRecord mixin to add conventions for flagging records as discarded.

It's the recommended gem by the Paranoia core.

> `paranoia` has some surprising behaviour (like overriding ActiveRecord's `delete` and `destroy`) and is not recommended for new projects. See [`discard`'s README](https://github.com/jhawthorn/discard#why-not-paranoia-or-acts_as_paranoid) for more details.

---


markets:
Thanks mate! :rocket:

PR #905, opened 2019-03-13
Add webpush

zaru:
## Project

[webpush](https://github.com/zaru/webpush)
[webpush | Rubygems](https://rubygems.org/gems/webpush/)

## What is this Ruby project?

This gem makes it possible to send push messages to web browsers from Ruby backends using the Web Push Protocol. It supports Message Encryption for Web Push to send messages securely from server to user agent.

I have not seen other implementations as far as I know.

---


markets:
Very cool project :ok_hand: 

Thanks for sharing!

PR #911, opened 2019-03-16
Add IRuby

kojix2:
## IRuby
https://github.com/SciRuby/iruby

## What is this Ruby project?
Ruby kernel for Jupyter. 
Jupyter is considered to be a standard tool for data processing. 

There are no similar projects as far as I know. 
Other Ruby kernel implementations may exist, but It may safely be said that they are not popular.

---


markets:
Thanks a lot @kojix2 for helping maintaining these two categories: Scientific and Machine learning! I'm not really involved in this area, so any kind of help here is really really appreciated!

PR #913, opened 2019-03-22
Ruby on Rails E-Commerce

nthompson196:
## Project

RoR E-Commerce. Demo Included

## What is this Ruby project?

We built an ecomm site using RoR and Refinery CMS. It includes a guide on how we did it, code snippets, and a demo.  

This focuses specifically on RoR for e-commerce. Great for entrepreneurial programmers working with Ruby. 

---


markets:
Hi @nthompson196 👋  

I think Awesome Ruby isn't the proper place for those kind of links. If we start adding links to blog-posts showing how to use a combo of gems, this list will become more useless, as it will eventually grow too much and we'll lose focus. On the other hand, I think we can't consider it a very well-known, popular and used (as per contrib guides) project in the Ruby ecosystem. 

So, thanks a lot for sharing here but I'm going to close this one.

PR #914, opened 2019-03-28
Add pry-rails to debugging tools

hieunguyentrung:
## Project

`pry-rails`
https://github.com/rweng/pry-rails

## What is this Ruby project?

Avoid repeating yourself, use pry-rails instead of copying the initializer to every rails project. This is a small gem which causes rails console to open pry. It therefore depends on pry.

Useful and bring pry to rails console

### DIFF
--- a/README.md
+++ b/README.md
@@ -461,10 +461,10 @@ Best suited for map-reduce or e.g. parallel downloads/uploads.
 
 ## Debugging Tools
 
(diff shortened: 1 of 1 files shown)


markets:
Hi @hieunguyentrung and thanks for sharing! Could you please add it in alphabetical order as per contrib guides?

hieunguyentrung:
Hi @markets, please check it again. Thanks.

PR #919, opened 2019-04-18
Add red chainer

kojix2:
## Red Chainer
https://github.com/red-data-tools/red-chainer

## What is this Ruby project?
This is a deep learning library written in pure Ruby.
Red Chainer can use GPU. 

This is a difficult question.

Red Chainer has no advantage over Deep Learning libraries written in Python or C.

Most of Ruby's deep learning libraries are bindings. Therefore, comparing with Red Chainer is not very appropriate. 
https://github.com/somaticio/tensorflow.rb

This project seems to have stayed at the experimental level.
https://github.com/jedld/tensor_stream

Yes, Red Chainer is written in pure Ruby. It runs fast on the GPU. It actually works.

However, from my point of view, the API is not Ruby-like enough. This is because Red Chainer is a clone library of Chainer. In Japan, Chainer is somewhat popular. But, Chainer is not a widely used library in the world. This also makes it difficult to evaluate Red Chainer.

The only library that can be compared with Red Chainer in a true sense is ruby-dnn. 
https://github.com/unagiootoro/ruby-dnn
API is Ruby-like and cute. However, ruby-dnn is developed as a personal hobby.  

* RubyKaigi2019
  * Red Chainer and Cumo: Practical Deep Learning in Ruby 
  * https://rubykaigi.org/2019/presentations/sonots.html#apr20

---


kojix2:
Well. I think I should try other Deep Learning libraries (especially bindings) more. 
Red Chainer is written in Ruby, but that is not enough to be awesome for everyone. 
I will close it once.

PR #924, opened 2019-05-14
Add render_async to Optimizations category

nikolalsvk:
### [render_async](https://github.com/renderedtext/render_async)

Blog posts: 
- https://semaphoreci.com/blog/2017/06/08/speeding-up-rails-pages-with-render-async.html
- https://medium.com/@nikolalsvk/rails-html-polling-without-javascript-dc28200d2a71

## What is this Ruby project?

render_async renders partials to your views asynchronously. This is done through adding JavaScript code that does AJAX request to your controller which then renders your partial into a Rails view.

It supports many customization options, polling, caching, retry partial load after failure, error handling.

It uses simple JavaScript to fetch partials compared to [get_schiwfty](https://github.com/danielwestendorf/get_schwifty) which is based on ActionCable

---

### DIFF
--- a/README.md
+++ b/README.md
@@ -825,7 +825,6 @@ Best suited for map-reduce or e.g. parallel downloads/uploads.
 
 * [bootsnap](https://github.com/Shopify/bootsnap) - Boot large Ruby/Rails apps faster.
 * [fast_blank](https://github.com/SamSaffron/fast_blank) - Provides a C-optimized method for determining if a string is blank.
-* [render_async](https://github.com/renderedtext/render_async) - Render partials to your views asynchronously and increase load performance of your pages.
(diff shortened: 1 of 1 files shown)


markets:
Hi @nikolalsvk and thanks a lot for contributing! 👍 

Nice catch! I'm not sure about the category, but I'm totally +1 in this one. Maybe it fits better under [View helpers](https://github.com/markets/awesome-ruby#view-helpers)? I'm saying this, because the gem does not perform an app optimization "per se", this can be an specific use-case from the app logic itself resolved with AJAX instead of other solutions.

markets:
Reminds me this one: https://github.com/antulik/pagelet_rails

nikolalsvk:
Thanks for commenting @markets!

View helpers category sounds good, I'll edit the PR and move it there. 

Oh yeah, it is similar to pagelet_rails, too sad that project is no more active :( render_async could use some features pagelet_rails has.

PR #933, opened 2019-07-25
Add fast_underscore

kddeisz:
## Project

fast_underscore

* [github](https://github.com/kddeisz/fast_underscore)
* [rubygems](https://rubygems.org/gems/fast_underscore)

## What is this Ruby project?

A C-extension for the `String#underscore` method from Rails.

I don't believe that are other similar ones besides `fast_blank`. This is `fast_blank` but for underscore.


markets:
Hi again @kddeisz 👋  

We usually don't consider this kind of resources for this list, like for example we don't list the `fast_blank` or gems alike.

kddeisz:
Oh - I was putting this directly below the `fast_blank` gem, which is why I thought I'd put it in here

markets:
ooh you're right sorry!

kddeisz:
@markets bumping this, it's at 23K at the moment.

PR #934, opened 2019-07-25
Add humidifier

kddeisz:
## Project

humidifier

* [github](https://github.com/kddeisz/humidifier)
* [rubygems](https://rubygems.org/gems/humidifier)

## What is this Ruby project?

A ruby CLI/library for interacting with the AWS CloudFormation API.

There are a couple of other CloudFormation ruby libraries (like SparkleFormation). In general most of these libraries act as a fancy JSON generator. This library instead gives you domain objects to represent AWS resources.


markets:
Cool one 🚀 thanks!

PR #935, opened 2019-07-25
Add prettier

kddeisz:
## Project

prettier

* [github](https://github.com/prettier/plugin-ruby)
* [rubygems](https://rubygems.org/gems/prettier)

## What is this Ruby project?

A automatic code formatter for Ruby.

Similar ones include rubyfmt, rufo, or some aspects of rubocop. The main difference is this uses prettier's main project so it gets the benefits of using their formatter.

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,6 +31,7 @@ Thanks to all [contributors](https://github.com/markets/awesome-ruby/graphs/cont
   * [Cloud](#cloud)
   * [CMS](#cms)
   * [Code Analysis and Metrics](#code-analysis-and-metrics)
+  * [Code Formatting](#code-formatting)
   * [Code Highlighting](#code-highlighting)
   * [Coding Style Guides](#coding-style-guides)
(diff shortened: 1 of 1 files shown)


markets:
Cool project @kddeisz 👌 still too fresh for this list (under 5k downloads), but promising.

Thanks for sharing!

kddeisz:
Yeah no worries. Just curious what's the threshold at which you accept it? I didn't see anything in the contributing doc.

markets:
We don't really have set a rule yet, it's hard... sometimes numbers only aren't a good indicator of "awesomeness", but for reference, in my mind I have a number: +10k downloads.

Maybe we should add this info to contribution guides?

In the meantime, we can leave it open, occasionally I review all opened issues and people can comment, +1 ...

kddeisz:
Yeah no worries. The thing with prettier is that it's also a JS package, so most of the downloads have been from that

markets:
> it's also a JS package, so most of the downloads have been from that

:point_up: that's relevant point I think :rocket:

PR #937, opened 2019-07-25
Update Rails version target for Refinery CMS

parndt:
Just updating the Rails version number that Refinery CMS supports :smile:


markets:
Thanks!

parndt:
Thanks to you too @markets 	!

PR #944, opened 2019-08-08
Add arli gem (a CLI-based Arduino Library Manager and new project generator)

kigster:
## Project

**Arli — Arduino Library Manager, searcher and installer.**

## What is this Ruby project?

**Arli** is an Arduino library manager with the ability to search libraries, install a single library, or bundle any number of libraries with the project. Arli also is a new project generator based on the [`arduino-cmake`](https://github.com/arduino-cmake/arduino-cmake) build system.

Arli is a rather simple and easy to use command-line tool which offers several indispensable features that help with Arduino project development, in particular for much larger projects with many dependencies and external libraries.

What's more, is that projects generated by Arli's generate command are highly portable from one system to another. Anyone can download your project build/upload with very little work.

[asciicast](https://asciinema.org/a/155289)

There are no similar Ruby projects, as far as I am aware. 

There is [`platformio`](https://platformio.org) written in Python, and there is a built-in [Arduino Library Manager](https://learn.adafruit.com/adafruit-all-about-arduino-libraries-install-use/library-manager), but nothing in Ruby.

---


kigster:
Thanks!!

PR #951, opened 2019-09-02
Add AssociateRB

lucasqueiroz:
### AssociateRB
- [GitHub](https://github.com/lucasqueiroz/associate-rb)
- [RubyGems](https://rubygems.org/gems/associate_rb)

## What is this Ruby project?

AssociateRB implements Kotlin's `associate` methods to Ruby's `Array`s. It allows you to easily convert an Array into a Hash.

- `Finishing Moves` - Although Finishing Moves includes some methods to convert Arrays into Hashes, it doesn't give you the option to manipulate the Keys and Values at the same time, or add the result to an existing Hash.

---


markets:
Hi @lucasqueiroz and thanks for sharing your library! Still too fresh for this list, only ~200 downloads according RubyGems (as a reference, we usually take it into account since 10k downloads). Thanks.

PR #952, opened 2019-09-16
Remove Go Package: Gitrob

igbanam:
Since this isn't an addition, I don't know how to follow the pull request template on this one.

I was looking through and saw that Gitrob — albeit being an interesting idea for a package — is not a Ruby tool.

Since this is awesome Ruby, I thought we should also curate as much as we collect


markets:
Hi @igbanam and thanks for contributing here! 👍 

Agreed, it was originally written in Ruby, but rewritten in Go ... https://twitter.com/michenriksen/status/1005027807183949824

PR #957, opened 2019-10-24
Add Zeitwerk

karlcoelho:
## Project

Zeitwerk

https://github.com/fxn/zeitwerk

## What is this Ruby project?

An efficient and thread-safe code loader for Ruby.

No similar ones. This is new.


markets:
Of course :ok_hand: 

Thanks!!

PR #970, opened 2020-02-03
add actionview-component

joelhawksley:
## Project

https://github.com/github/actionview-component
https://www.ruby-toolbox.com/projects/actionview-component
https://rubygems.org/gems/actionview-component/

## What is this Ruby project?

ActionView::Component is a framework for building view components in Rails.

ActionView::Components are Ruby classes that are used to render views. They take data as input and return output-safe HTML. Think of them as an evolution of the presenter/decorator/view model pattern, inspired by React Components.

ActionView::Component is similar to other view component libraries, but is specifically designed for tight conceptual integration into Rails. It is an extraction from the GitHub monolith.

---

### DIFF
--- a/README.md
+++ b/README.md
@@ -125,6 +125,7 @@ Thanks to all [contributors](https://github.com/markets/awesome-ruby/graphs/cont
 ## Abstraction
 
 * [ActiveInteraction](https://github.com/orgsync/active_interaction) - Manage application specific business logic.
+* [ActionView::Component](https://github.com/github/actionview-component) - View components for Rails.
 * [Apotomo](https://github.com/apotonick/apotomo) - Based on Cells, Apotomo gives you widgets and encapsulation, bubbling events, AJAX page updates, rock-solid testing and more.
 * [Cells](https://github.com/trailblazer/cells) - View Components for Rails.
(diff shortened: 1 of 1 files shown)


markets:
Hi @joelhawksley and thanks a lot for contributing here 👍 

Maybe I'm wrong, but I think View components were merged to Rails, or at least, I remember an open PR with quite similar feature.

joelhawksley:
@markets that PR, https://github.com/rails/rails/pull/36388, only had preliminary support (I'm the author if it 😄), enabling libraries such as ActionView::Component to hook into Rails.

`actionview-component` is the library itself, where we're working on what we plan to upstream into Rails eventually.

markets:
Ah yes! I'm speaking about that pull request 👍 

I'm a bit confused now 🤔... so Rails master has preliminary "support", but not ready to use Components? Or Components can be used, but this library adds more features?

joelhawksley:
Rails master has the hooks necessary to use the `actionview-component` gem, which provides the `ActionView::Component` base class for building view components.

markets:
ok, understood, I just made a quick look at the PR diff 👍 

Nice project! would be amazing to merge this upstream 👌

joelhawksley:
Closing this PR as we are planning to rename the gem.

PR #987, opened 2020-04-25
Remove cane

matt-harvey:
cane has not had a commit for four years; and the [README](https://github.com/square/cane) recommends using Rubocop instead of cane for new projects.


markets:
Thank you @matt-harvey 👍

PR #1000, opened 2020-06-05
Simple Scheduler

bmedenwald:
## Project

[Simple Scheduler](https://github.com/simplymadeapps/simple_scheduler), an enhancement for Heroku Scheduler + Sidekiq for scheduling jobs at specific times. [Check out our intro blog post to learn more](http://www.simplymadeapps.com/blog/2017/01/simple-scheduler-schedule-recurring-tasks-to-run-at-any-interval/).

## What is this Ruby project?

Simple Scheduler is a scheduling add-on that is designed to be used with [Sidekiq](http://sidekiq.org) and [Heroku Scheduler](https://elements.heroku.com/addons/scheduler). It gives you the ability to **schedule tasks at any interval** without adding a clock process. Heroku Scheduler only allows you to schedule tasks every 10 minutes, every hour, or every day.

Every option we evaluated seems to have the same flaw: **If your server is down, your job won't run.** With Simple Scheduler, we use Sidekiq's robust scheduling and retrying system to make sure your schedule jobs run, even if the server was down at the moment it should have ran.

Everything is driven from a simple YML file for readability and developer happiness. No crontab knowledge required. ActiveSupport durations FTW!

---

### DIFF
--- a/README.md
+++ b/README.md
@@ -1016,6 +1016,7 @@ Best suited for map-reduce or e.g. parallel downloads/uploads.
 * [resque-scheduler](https://github.com/resque/resque-scheduler) - A light-weight job scheduling system built on top of Resque.
 * [rufus-scheduler](https://github.com/jmettraux/rufus-scheduler) - Job scheduler for Ruby (at, cron, in and every jobs).
 * [Sidekiq-Cron](https://github.com/ondrejbartas/sidekiq-cron) - A scheduling add-on for Sidekiq.
+* [Simple Scheduler](https://github.com/simplymadeapps/simple_scheduler) - An enhancement for Heroku Scheduler + Sidekiq for scheduling jobs at specific times with a readable YML file.
 * [Whenever](https://github.com/javan/whenever) - A Ruby gem that provides a clear syntax for writing and deploying cron jobs.
 
 ## Scientific


markets:
Hi @bmedenwald and thanks for sharing this! Seems a nice and well maintained project 👍 

Can I ask what is the main difference, for example, compared to Sidekiq-Cron?

bmedenwald:
@markets Sure can.

While they both use Sidekiq, the biggest difference IMO would be the readability. You don't need to understand cron and how the asterisk notation works. Simple Scheduler utilizes ActiveSupport's durations for a readable file. Here's an example:

```yml
# Global configuration options. The `queue_ahead` and `tz` options can also be set on each task.
queue_ahead: 360 # Number of minutes to queue jobs into the future
queue_name: "default" # The Sidekiq queue name used by SimpleScheduler::FutureJob
tz: "America/Chicago" # The application time zone will be used by default if not set

# Runs once every 2 minutes starting at the top of the hour
simple_task:
  class: "SomeActiveJob"
  every: "2.minutes"
  at: "*:00"

# Runs once every day at 4:00 AM. The job will expire after 23 hours, which means the
# job will not run if 23 hours passes (server downtime) before the job is actually run
overnight_task:
  class: "SomeSidekiqWorker"
  every: "1.day"
  at: "4:00"
  expires_after: "23.hours"

# Runs once every half hour, starting on the 30 min mark
half_hour_task:
  class: "HalfHourTask"
  every: "30.minutes"
  at: "*:30"

# Runs once every week on Saturdays at 12:00 AM
weekly_task:
  class: "WeeklyJob"
  every: "1.week"
  at: "Sat 0:00"
  tz: "America/New_York"
```

It's easy as pie to reason about when a job should be running.

While Heroku is not required (just need to periodically run a rake task), this was geared in the beginning to make that a drop-in replacement for their more-complicated clock process setup (which is bitten by your server going down).

markets:
Oh I see, that's a nice api, it reminds me the one used by [javan/whenever](https://github.com/javan/whenever).

I used this one in the past too: https://github.com/moove-it/sidekiq-scheduler. It also allows some kind of friendly api, instead of cron * notation only. But I'm using `sidekiq-cron` lately in all my Sidekiq projects. It worked pretty good so far for me in several projects, has less deps and since we also use this in Sinatra apps, we don't need to have the Rails gem as runtime dependency 😬

bmedenwald:
Sure! Yes, we're the authors of this gem and having a nice API was Job 1. We used Readme Driven Development, and after we were happy with the Readme and knew we'd use this gem if it existed, then we went on and built it :)

bmedenwald:
Revisiting this: the HUGE difference between Simple Scheduler and whenever is replay-ability. If your computer is down when the cronjob should run, it will never run. That's the flaw in whenever and most other projects. With Simple Scheduler, we piggy-back off of Sidekiq to make sure that even if your job doesn't run immediately, it will get picked up and run again when Sidekiq gets to it. 

We provide the original run time, so you can even act on that information to determine how long after the time it should be Ok to run and when it should throw you an error message so you can investigate further. This allows the [developer to decide in case of an outrage if an old job should run or not](https://github.com/simplymadeapps/simple_scheduler#handling-expired-jobs).

PR #1001, opened 2020-06-13
Updated to Git Lint 1.0.0

bkuhlmann:
## Project

Git Lint is the official port of the original Git Cop project that
avoids references to violence or violent terminology that might make
anyone feel unwelcome. The name change was necessary in order to better
support the [Black Lives Matter](https://blacklivesmatter.com) and
[Defund Police](https://8cantwait.org) movements.

## What is this Ruby project?

This is a replacement/update to the original Git Cop link.

No difference in features between Git Cop and Git Lint. Git Lint is a direct port but with a better name.

---


bkuhlmann:
I wanted to check back and see if there were any issues with this pull request? This is only a name changes and update to the original documentation so wasn't thinking this would take long to review. If there is anything else I can do to push this forward, let me know. :wink:

PR #1003, opened 2020-07-10
Add puma-pool-usage gem.

bmedenwald:
## Project

[Puma Pool Usage](https://github.com/simplymadeapps/puma-pool-usage), a Puma plugin that interprets usage statistics for the percentage of your resources under load. Uses Rails to log this statistic. 

## What is this Ruby project?

Inspired by [Tomas Ruzicka](https://github.com/LeZuse) and Heroku's pool usage statistics. Knowing how much of your Puma pool is busy and how much you have available is critical to knowing whether you have enough resources available to handle increases in web server load. 

Look at this neat log:

```
source=PUMA pid=74840 sample#pool_usage=0.8
```

If your web server is fielding no requests, usage will be 0.0 (0%). If every single resource has a request, usage will be 1.0 (100%). If you have a backlog of requests that cannot be processed because no resources are available, usage can be over 100% (like 1.4).

Having this data allows anyone to chart this data for performance, setup alerts, or scale your servers based on it. These are all great idea, but it's beyond the scope of this plugin. This plugin provides the data, what you do with the data is up to you.

Puma pool statistics are provided via the Puma internals. This plugin surfaces them in your logs with an algorithm that takes into account your Puma settings and cluster mode. The closest similar project would be the [barnes gem](https://github.com/heroku/barnes) that does more statistical analysis but works exclusively with Heroku. This gem works on any host running Puma.

---

### DIFF
--- a/README.md
+++ b/README.md
@@ -955,6 +955,7 @@ Best suited for map-reduce or e.g. parallel downloads/uploads.
 * [bullet](https://github.com/flyerhzm/bullet) - Help to kill N+1 queries and unused eager loading.
 * [Derailed Benchmarks](https://github.com/schneems/derailed_benchmarks) - A series of things you can use to benchmark any Rack based app.
 * [Peek](https://github.com/peek/peek) - Visual status bar showing Rails performance.
+* [puma-pool-usage](https://github.com/simplymadeapps/puma-pool-usage) - Puma plugin to calculate and log your web server pool usage.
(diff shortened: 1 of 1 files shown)


markets:
Hi @bmedenwald and thanks for sharing! 

This project is still too fresh for this list. As per contribution guides, we consider projects over 10k downloads  and according to RubyGems (https://rubygems.org/gems/puma-pool-usage) this is has 390. Which, more or less, means it has some relevance in the Ruby ecosystem.

bmedenwald:
Roger that. Will wait for the future :)

PR #1005, opened 2020-08-17
Add blueprinter as API builder library

diepsohung:
## Project

- [Github](https://github.com/procore/blueprinter)
- [RubyGems](https://rubygems.org/gems/blueprinter/versions/0.6.0)

## What is this Ruby project?

Blueprinter is a JSON Object Presenter for Ruby that takes business objects and breaks them down into simple hashes and serializes them to JSON. It can be used in Rails in place of other serializers (like JBuilder or ActiveModelSerializers). It is designed to be simple, direct, and performant.

It heavily relies on the idea of views which, similar to Rails views, are ways of predefining output for data in different contexts.

---


markets:
Thanks @diepsohung for sharing this!

PR #1020, opened 2020-11-30
Add ruby-spellchecker

omohokcoj:
## Project

ruby-spellchecker [https://github.com/omohokcoj/ruby-spellchecker](https://github.com/omohokcoj/ruby-spellchecker)
RubyGems: https://rubygems.org/gems/ruby-spellchecker

## What is this Ruby project?

It helps to eliminate common spelling and grammar errors from the text (to prepare it for NLP for instance).

Other ruby spellchecking gems are quite outdated and not maintained. Many of them use hunspell-ffi to check all words from the text against existing English dictionaries and to suggest corrections based on jaro/levenshtein distance. Hunspell's approach can produce many false positives and requires shared libraries to be installed. This gem uses a large list of common errors collected from various open sources so it doesn't produce false positives and can be used for autocorrection. Also, it's written in pure ruby.


markets:
Thanks for sharing @omohokcoj! 

Still too fresh for this list (around 300 downloads), but cool project 👌🏼. As per contrib guides, we usually consider new additions when they reach ~10k downloads.

dohungthinhSTS:
@omohokcoj That's a good job man!!

PR #1022, opened 2020-12-03
Add attr-gather gem to list

ianks:
## Project

Name: `attr-gather`
GitHub: https://github.com/ianks/attr-gather
RubyGems: https://rubygems.org/gems/attr-gather

## What is this Ruby project?

[attr-gather](https://github.com/ianks/attr-gather) - A gem for creating workflows that "enhance" entities with extra attributes. At a high level, [attr-gather](https://github.com/ianks/attr-gather) provides a process to fetch information from many data sources (such as third party APIs, legacy databases, etc.) in a fully parallelized fashion.

- Offers DSL for defining workflows and merging the results from each task
- Execution engine optimally parallelizes the execution of the workflow dependency graph
- Easily testable
- Ability to filter out bad data using schemas with [dry-validation](https://dry-rb.org/gems/dry-validation)

- Operates on a single entity rather than a list, so easily can be adopted in existing systems
- Focuses on the "fetching" and filtering of data solely, and not transformation or storage
- Fully parallel execution engine
- Focuses on having a clean PORO interface to make testing simple
- Provides a declarative interface for merging results from many sources (APIs, legacy databases, etc.) which allows for prioritization

---

### DIFF
--- a/README.md
+++ b/README.md
@@ -412,6 +412,7 @@ Best suited for map-reduce or e.g. parallel downloads/uploads.
 
 ## Data Processing and ETL
 
+* [attr-gather](https://github.com/ianks/attr-gather) - A gem for creating workflows that "enhance" entities with extra attributes. At a high level, attr-gather provides a process to fetch information from many data sources (such as third party APIs, legacy databases, etc.) in a fully parallelized fashion.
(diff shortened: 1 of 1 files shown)


markets:
Thanks for sharing @ianks!

Still too fresh for this list, but cool project 👌🏼. As per contrib guides, we usually consider new additions when they reach ~10k downloads.

ianks:
Awesome! Should I keep this PR open until that benchmark is hit?

PR #1030, opened 2021-03-04
Add Active Entry gem to authentication and authorization section.

0x7466:
## Project

Active Entry - https://github.com/TFM-Agency/active_entry

## What is this Ruby project?

Active Entry is a simple and secure authentication and authorization system for your Rails application, which lets you to authenticate and authorize directly in your controllers.

The problem I had with other auth gems was, that they mostly needed Policies, Abilities, etc. I just wanted a safe and simple way to authenticate and authorize. Active Entry was born. :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -207,6 +207,7 @@ Thanks to all [contributors](https://github.com/markets/awesome-ruby/graphs/cont
 * [Shield](https://github.com/cyx/shield) - Authentication protocol for use in your routing and model context.
 * [Sorcery](https://github.com/Sorcery/sorcery) - Magical Authentication for Rails 4, and Rails 5.
 * [warden](https://github.com/hassox/warden) - General Rack Authentication Framework.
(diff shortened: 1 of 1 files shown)


markets:
Hi @tobiasfeistmantl and thanks for sharing! As per [contribution guides](https://github.com/markets/awesome-ruby/blob/master/CONTRIBUTING.md), this is too fresh (only 300 downloads, min: 20k). Let's keep it open and see how the project evolves.

Let me add a couple of comments more 😄:
- we list projects only in 1 category (in that case, I think it should eventually go under Authorization, as this gem doesn't seem to perform authentication stuff).
- the explanation under the "differences between similars" question seems a bit subjetive, IMHO having separated "policies" or "abilities" is a better approach since it doesn't mess up with all the logic in the controllers and you can test them easily.

Best,
Marc

PR #1036, opened 2021-05-26
Add ruby-clock

jjb:
## Project

Source: https://github.com/jjb/ruby-clock
RubyGems: https://rubygems.org/gems/ruby-clock

## What is this Ruby project?

a scheduler which can replace cron. runs jobs in a persistent process, each in their own thread

1. based on venerable rufus for scheduling mechanics
2. runs jobs in persistent process
3. easy to configure to be a good citizen running in environments like heroku


markets:
I was checking your project @jjb and it seems you're close to a new v2 version (now in beta). After you ship this new release, I think we can consider merging this PR.

jjb:
@markets Okay, 2.0.0 is out!

PR #1056, opened 2021-11-05
Updating httpx description

HoneyryderChuck:
The current description was particularly inaccurate, and makes it look like httpx is an extension to httprb, which is not (it's its own gem).


markets:
Thanks for the update @HoneyryderChuck!

markets:
For the sake of reference, it was initially merged in: https://github.com/markets/awesome-ruby/pull/820

HoneyryderChuck:
🙏  thx @markets

PR #1078, opened 2022-08-10
Add Ruby@kandi

kandikits:
Hello - request review and approval on the inclusion of the following on your list under section "Discover"
It would be a good addition for this interesting article. Would you like to include it in your list? We'd love to hear your feedback.
Thanks!

* [Explore Ruby @kandi](https://kandi.openweaver.com/explore/ruby) - Discover & find a curated list of popular & new Ruby libraries, top authors, trending project kits, discussions, tutorials & learning resources.

### DIFF
--- a/README.md
+++ b/README.md
@@ -544,6 +544,7 @@ Best suited for map-reduce or e.g. parallel downloads/uploads.
 Where to discover new Ruby libraries, projects and trends.
 
 * [Awesome Ruby @LibHunt](https://ruby.libhunt.com) - Your go-to Ruby Toolbox. The Awesome Ruby collection + ordering by popularity, additional metadata and comparisons.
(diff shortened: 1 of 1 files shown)


markets:
Hey @kandikits and thanks for sharing this resource.

Unfortunately, I'm not going to merge it. Just took a look around it and I don't see so much value on the information presented. Or in some cases, "weird" information... for example, when you enter that page, the 1st info you can see is:

Captura de pantalla 2022-08-30 a las 22 03 46

Both projects, Rails and Jekyll, have new (and recommended) releases... 🤔 sorry I can't understand why those are marked as the "Popular New Releases". 

Then, some sections below:

Captura de pantalla 2022-08-30 a las 22 08 31

I wouldn't say those are the "Top Authors" in the Ruby ecosystem.

I don't know what's running under the hood 😅, but it seems they aggregate some data from different APIs, without any "manual curation" or "post-process", which results in some weird info...

PR #1098, opened 2023-03-16
Update README.md

bradgessler:
Add Sitepress to CMS

## Project

https://sitepress.cc
https://github.com/sitepress/sitepress

## What is this Ruby project?

Manage content files in Rails with the same features you'd expect from static site generators like Jekyll, Middleman, and Bridgetown.

### Frontmatter
Add metadata to the top of your content page files that you can plug into your templates, collections, and more.

### Globbing
Quickly generate lists, navigation menus, collections, and more with Dir#glob.

### Page models
Wrap pages in custom Ruby classes for even more powerful ways of traversing your sites content for navigation, grids, or collections.
### Layouts, partials, & pages
Sitepress is built on top of Rails, so you will feel right at home rendering pages, layouts, & partials.

### Templating languages
Markdown, Slim, HAML, and Erb. All the templating libraries that work with Rails will work with Sitepress, and that's a lot of stuff.

### Static compiler
Sitepress also works as a static site compiler so you can eject the website out of your Rails app when you're ready to start breaking apart the monolith.

Most CMSs in Rails require databases, which isn't a great way to manage content, esp if it needs to be synced between environments. Sitepress uses the mighty file and git to manage content.

This CMS also works without Rails, like Jekyll, Middleman, and Bridgetown. It also works great inside of Rails, and is just as powerful as those static site compilers. You can actually move a Sitepress project in and out of Rails as your project requirements change.

---

### DIFF
--- a/README.md
+++ b/README.md
@@ -302,6 +302,7 @@ Thanks to all [contributors](https://github.com/markets/awesome-ruby/graphs/cont
 * [PushType](http://www.pushtype.org/) - A next generation, open source content management system for Ruby on Rails.
 * [Radiant](http://radiantcms.org) - A no-fluff, open source content management system designed for small teams.
 * [Refinery CMS](https://www.refinerycms.com) - An open source Ruby on Rails content management system for Rails 5.2+.
+* [Sitepress](https://sitepress.cc/) - Manage content files in Rails with the same features you'd expect from static site generators like Jekyll, Middleman, and Bridgetown.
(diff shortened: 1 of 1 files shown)


markets:
> I've got a data feature in the works for Sitepress 4.x. It will have a ./data/*.yml folder that works like MM

🎉 Amazing! 🎉 That's a really nice feature I missed from MM.

PR #1114, opened 2023-05-15
Replace Raven Ruby with Sentry Ruby

st0012:
## Project

- https://github.com/getsentry/sentry-ruby
- https://rubygems.org/gems/sentry-ruby/

## What is this Ruby project?

It's the new Sentry SDK for Ruby, which replaces the old Raven Ruby library.

- According to the latest [Ruby and Rails Community Survey](https://rails-hosting.com/2022/#which-error-tracking-tools-do-you-use-in-production), Sentry is the most popular option for error monitoring. So I think it's worth being included in the section.
- The old Raven Ruby has been deprecated for 3 years, and the new `sentry-ruby` gem now has been downloaded for more than 27 million times ([link](https://rubygems.org/gems/sentry-ruby/)).

---


markets:
Thanks @st0012!

st0012:
@markets Thanks for the prompt response!

PR #1133, opened 2024-03-22
Add `pluck_in_batches`

fatkodima:
## Project

https://github.com/fatkodima/pluck_in_batches
https://rubygems.org/gems/pluck_in_batches

## What is this Ruby project?

A faster alternative to the custom use of `in_batches` with `pluck`.


markets:
Thanks @fatkodima 🚀

PR #1144, opened 2024-08-15
add Even Better Specs

glaucocustodio:
## Even Better Specs

https://evenbetterspecs.github.io/
https://github.com/evenbetterspecs/evenbetterspecs.github.io

## What is this Ruby project?

Even Better Specs is an opinionated set of best practices to support the creation of tests that are easy to read and maintain.

There are a lot of differences, Even Better Specs for example do not recommend the usage of `let` variables neither `shared_examples`


markets:
Hey @glaucocustodio thanks for this!

I think we should wait some time for this merge... The original resource is quite well-known (it's there for ages) in the community and since these changes are quite opinionated, it seems a good idea to wait for some metrics and your guide to take some relevant position in the ecosystem.

PR #1145, opened 2024-08-18
Add `AnnotateRb` gem

drwl:
## Project

Adds `AnnotateRb` gem to the list of gems.

[Rubygems](https://rubygems.org/gems/annotaterb)
[Github repo](https://github.com/drwl/annotaterb)

## What is this Ruby project?

It's an active and maintained hard fork of the old Annotate gem. It improves the developer experience by adding model database schemas as text comments in corresponding ActiveRecord models. It also supports annotating routes.

It's actively maintained compared to it's predecessor, the Annotate gem.

### DIFF
--- a/README.md
+++ b/README.md
@@ -588,6 +588,7 @@ Where to discover new Ruby libraries, projects and trends.
 ## Documentation
 
 * [Annotate](https://github.com/ctran/annotate_models) - Add a comment documenting the current schema to the top or bottom of each of your ActiveRecord models.
+* [AnnotateRb](https://github.com/drwl/annotaterb) - Adds database schema annotations for your ActiveRecord models as text comments as well as routes. An active and maintained hard fork of Annotate. 
 * [Apipie](https://github.com/Apipie/apipie-rails) - Rails API documentation and display tool using Ruby syntax.
 * [Asciidoctor](https://asciidoctor.org) - A fast, Ruby-based text processor & publishing toolchain for converting AsciiDoc to HTML5, DocBook, EPUB3, PDF & more.
 * [Documentation](https://github.com/adamcooke/documentation) - A Rails engine to provide the ability to add documentation to a Rails application.
--- a/README.md
+++ b/README.md
@@ -587,7 +587,6 @@ Where to discover new Ruby libraries, projects and trends.
 
 ## Documentation
 
-* [Annotate](https://github.com/ctran/annotate_models) - Add a comment documenting the current schema to the top or bottom of each of your ActiveRecord models.
 * [AnnotateRb](https://github.com/drwl/annotaterb) - Adds database schema annotations for your ActiveRecord models as text comments as well as routes. An active and maintained hard fork of Annotate. 
 * [Apipie](https://github.com/Apipie/apipie-rails) - Rails API documentation and display tool using Ruby syntax.
 * [Asciidoctor](https://asciidoctor.org) - A fast, Ruby-based text processor & publishing toolchain for converting AsciiDoc to HTML5, DocBook, EPUB3, PDF & more.


markets:
Hey @drwl 👋🏼 thanks for this!

Maybe we should replace (delete ✂️) the original repo too? Would be nice to wait for some official answer 👀 https://github.com/ctran/annotate_models/issues/1028.

drwl:
I don't know if/when we will get an official answer, there's https://github.com/ctran/annotate_models/issues/1001 from earlier this year with also no answer.

As to whether or not to replace the original repo, I believe that's your decision. I can adjust the PR to delete it if you want.

markets:
Yep, I've already seen that earlier issue too, but would be nice any kind of "official" movement/message from the original maintainers. For such relevant and used gems, it's much better any official communication to avoid confusions within the Ruby ecosystem (specially for new comers). Did you try to contact the original maintainer? If they don't want to maintain it anymore, at least a message in the README (pointing to your fork) would help a lot!

In terms of this PR, I'd also prefer to have only one resource, with the goal to make it clear for all the people. Having both links in this collection it seems that creates again confusion for users checking this list.

drwl:
> For such relevant and used gems, it's much better any official communication to avoid confusions within the Ruby ecosystem (specially for new comers).

Totally agree there. 

> Did you try to contact the original maintainer?

> If they don't want to maintain it anymore, at least a message in the README (pointing to your fork) would help a lot!

Yes I have tried contacting in the past and no response. Also, an even earlier Github thread/issue.

https://github.com/ctran/annotate_models/issues/913

> In terms of this PR, I'd also prefer to have only one resource, with the goal to make it clear for all the people.

I can modify the PR to delete the old one, let me know if that's what you mean.

markets:
Thanks for jumping in @marcoroth! I'm also using the original one in a couple of small apps (Rails 6.1 and 7.0) with no issues at all. But I didn't introduced DB changes for months... so I'm really not sure about the current status of the original repo.

Maybe it's better idea to wait for any kind of official communication before moving forward and merge this PR now? Would be nice to talk about it with current maintainers.

drwl:
> I think it's fair to say that it could be considered "done" from the maintainers POV.

Calling it "done" is an understandable opinion, and for most setups it is. The core functionality _should_ work, because all the gem really does is find model files and reflect on their ActiveRecord columns. I would say there are some outstanding usability bugs in the old gem like swallowing comments, but if you don't come across that in your workflow then there really isn't a need to upgrade at this time (note: future Rails versions/changes could change things).

>  @drwl have you contacted the RubyGems team to see if they can do something about it? I think they might be able to help with that, so we don't have to have forks, which are also very similarly named, which adds a lot of confusion in my opinion.

Just curious, has there been precedent about this? Afaik there's no imminent security risk of the old gem being up, so not sure why they would do anything about it. As an example, the old CanCan gem is still on Rubygems even though it was community forked years ago. My gem/fork also is a hardfork in that it uses a new namespace.

> Maybe it's better idea to wait for any kind of official communication before moving forward and merge this PR now? Would be nice to talk about it with the current maintainers.

Ultimately it's your decision regarding the awesome ruby list. Given that the oldest out of the 3 issues is 3 years old, I don't see anything changing in the coming years.

marcoroth:
@drwl I was more alluding to the fact that they might be able you get ownership over the gem so you could continue maintaining and releasing it. But I guess this already too late, since you have your fork and gem already.

drwl:
Ahh got it, yeah I changed the namespace (and the name) because there were some big changes that I wanted to make (e.g. removing config in Rakefile) and thought it made more sense. Also, to signify that there were big breaking changes between the old gem and annotaterb.

drwl:
I got an email from Github about the PR -- just checking is there anything needed on my end? The last PR in the old repo was merged in almost a year ago.

https://github.com/ctran/annotate_models/commits/develop/

markets:
Not really, I'm still not sure what's the best move here ...

- Having both resources seems like an exception and confusing for users checking this repo. I'd like to avoid this.
- Yours seems more up-to-date and with new features.
- But the original still works for me without any problem and it's like the canonical for most of the community.

So, I have no idea what's the best decision here. Let's leave it open for now, maybe some day we'll resolve these doubts.

markets:
@drwl Could you please delete the original project link? I think it's time to merge this one.

PR #1173, opened 2025-09-22
Add ChainMail to email tools list in README

taltas:
## Project
ChainMail
https://github.com/taltas/chain_mail

## What is this Ruby project?

ChainMail is a Ruby gem for reliable email delivery. It lets Rails apps send transactional emails through multiple providers with automatic failover, ensuring your messages get delivered even if one service fails.

One does not exist like this.


taltas:
Ah that's fair, hopefully it will get there soon. I'll close this for now.

PR #1216, opened 2026-06-02
Add leakferret to Security

missusk:
Adds [leakferret](https://github.com/leakferrethq/leakferret-ruby), an open-source, context-aware secret scanner available as a Ruby gem (`gem install leakferret`).

It finds hardcoded secrets, verifies which ones are actually live by calling the provider (rather than only pattern-matching), and rewrites the leak to read from ENV. It also ships a CLI and an MCP server. MIT licensed.

Placed alphabetically in the Security section.


markets:
Hey 👋🏼 please read the contrib guides first! Projects should be relevant (at least 30k downloads) within the Ruby ecosystem, this is not a place to promote new projects (for free, you can do it by sponsoring my account).

missusk:
totally fair, thanks for the quick reply! i didn't catch the 30k threshold in the guidelines, my bad. i'll circle back once it's past that. thanks for maintaining the list 🙏

PR #1222, opened 2026-07-01
Add xberg to Spreadsheets and Documents

Goldziher:
Add xberg to Spreadsheets and Documents. Ruby binding (RubyGems) over a Rust core, MIT. One link per PR per CONTRIBUTING.


markets:
Same as https://github.com/markets/awesome-ruby/pull/1223#issuecomment-4853178977

In this case, even worse... ~300 downloads: https://rubygems.org/gems/xberg

